PR #5, opened 2022-09-07
Removed child.css from repo + cosmetic bug fix in SLR + PHP warning with SPEI

perronld:


perronld:
I'm thinking about adding https://github.com/marketplace/actions/sass-build as a test for PR

perronld:
> I'm thinking about adding https://github.com/marketplace/actions/sass-build as a test for PR

This will not be completed in this current PR.

matprov (approved):

PR #6, opened 2022-11-07
H7 oct31

phil-evans:
- analyze page: new decimal places field (CLIM-255)
- header collapse bug CLIM-251)
- news posts: permalink (CLIM-237)
- news: author fields (CLIM-235)
- case studies: chart export bug (CLIM-231)

### DIFF
--- a/climate-data-ca/resources/js/analyze-functions.js
+++ b/climate-data-ca/resources/js/analyze-functions.js
@@ -402,7 +402,9 @@
 
         })
 
-
+				$('body').on('input', '.type-number :input', function() {
+					validate_inputs()
+				})
 
         function ChangeLayers(sector_value) {
             $('#lat').val("");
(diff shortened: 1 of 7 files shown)


perronld:
Please note that a pull request is not required to push to the demo branch... you can push directly. The pull request is required to push to the main branch (production).

I also notice that you are using the wrong parameter name for the decimals. (see: https://pavics-sdi.readthedocs.io/projects/finch/en/latest/processes.html )

I will perform the fix the and then the merge outside this PR.

climate-data-ca/resources/js/analyze-functions.js
```diff
@@ -1485,6 +1494,28 @@
                             this_has_val = true
                         }
                         break
+												
+										case 'number' :
```
perronld, line 1498: Project convention for indentation is 4 spaces, no tabs.

climate-data-ca/resources/js/analyze-functions.js
```diff
@@ -1485,6 +1494,28 @@
...
+										
+												var this_min = parseInt($(this).attr('min')),
```
perronld, line 1500: "var" define global javascript and can easily lead to unmaintable code. Please use "let" instead

PR #19, opened 2023-02-08
CLIM-308 new news section

phil-evans:


climate-data-ca/functions.php
```diff
@@ -162,6 +162,20 @@ function child_theme_enqueue()
         wp_enqueue_script('archive-functions');
 
     }
+    
+    if ( is_page ( 'news' ) ) {
```
perronld, line 166: This won't work in French

perronld (changes requested):
A quick look in French in demo shows that something doesn't work

climate-data-ca/functions.php
phil-evans: replaced `is_page` function with `is_page_template`

perronld (approved):

PR #20, opened 2023-02-13
Iframe feature

ahandan:
Adding a timeout to the Iframe, 

The timeout is used to for visibility on the user side of the url pulling.
This timeout is ensuring that any application will be mounted correctly in the time allowed by the user.

By default, no timeout is set

If the field timeout timer is used, then a listener will be setup and listening to events from the iframe waiting for this event message from the application, to cancel its timeout error function.


perronld:
Unfortunately, this PR doesn't make sense with the git history. Features branch must start from main branch, not demo branch (and demo must never be merged in the feature branch).

You'll have to start a new branch from main, and integrate your changes manually in it, then issue a new PR.

PR #22, opened 2023-02-27
CLIM-336 change chart options by abs/delta selection

phil-evans:


climate-data-ca/resources/js/child-functions.js
```diff
@@ -680,9 +680,23 @@ function displayChartData(data, varDetails, download_url, query, container) {
         }
     });
 
-    // 30y averages is by default, so trigger the change event on load
-    $('#chartoption2-' + query['var']).trigger('change');
-
+    if (
+        $('body').find('#toggle-b').length &&
```
perronld, line 684: While it works... this selector is not very descriptive for the reader.
Can we use  $('input[name="absolute_delta_switch"]:checked').val()  instead, since it's this one used elsewhere in the code?

perronld (changes requested):

climate-data-ca/resources/js/child-functions.js
perronld: I will take care of this one.
I will make sure this variable is part of "query"

PR #38, opened 2023-04-19
related content updates

perronld:


perronld:
@phil-evans An auto-format was done on single-variable.php, which added more than 1200 changes... of course there is are merge conflicts.

PR #64, opened 2023-10-23
CLIM-453 ignored any errors caused by Google Analytics code

perronld:
Only 3 significant lines.. make sure to hide whitespace when viewing the diff.

### DIFF
--- a/climate-data-ca/resources/js/child-functions.js
+++ b/climate-data-ca/resources/js/child-functions.js
@@ -359,37 +359,42 @@ function getGA4EventNameForVariableDownloadData(chartDataFormat, keySelected) {
 }
 
 function setDataLayerForChartData(chartDataFormat, chartData, query) {
-    if (!chartDataFormat.length) {
-        return;
-    }
-    let eventName = getGA4EventNameForVariableDownloadData(chartDataFormat, query['var']);
-    let overlayTitle = $('.overlay-title').text();
-
-    let addStr = "";
-    for (let index = 0; index < chartData.series.length; index++) {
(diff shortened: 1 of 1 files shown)


climate-data-ca/resources/js/child-functions.js
```diff
@@ -359,37 +359,42 @@ function getGA4EventNameForVariableDownloadData(chartDataFormat, keySelected) {
...
+            'event': eventName,
+            'chart_data_event_type': eventName,
+            'chart_data_settings': chartDataSettings,
+            'chart_data_columns': addStr,
+            'chart_data_dataset': query.dataset,
+            'chart_data_format': chartDataFormat,
+            'chart_data_view_by': variableDownloadDataVewBy
+        });
+    } catch (err) {  // we ignore any Google Analytics errors to keep the site functional for the user in case of problem
+        console.error(err);
```
francisPLT, line 395: Is there a more precise error we could catch?

Something like this:

```
...
} catch (err) {
    console.error(err);
    if (err.constructor !== "GoogleAnalyticsErrorTypeWhatever") {
        throw err
    }
}
```

If there's a precise error type we can catch that's only relevant to Google Analytics, it could be a nice and a small addition in order to still pick up other potentially relevant errors.

If not, then I say it's good enough considering V2 is on it's way.

perronld: CLIM-453 is caused by a null for mora. So doesn't worth to be fixed due to V2.

francisPLT (approved):

PR #71, opened 2023-11-07
H7 nov7 misc bugs

phil-evans:
fixed an error in functions.php that doesn't properly deregister 'font awesome' from the parent theme


fw-child/languages/fr_CA.mo
perronld, line 1: If I remember, .mo files are compiled from the .po?

PR #72, opened 2023-11-20
wiring up controls & layer swapping for map page

phil-evans:
big step forward for implementation of the map page, including:

- pulling visualization layers from API
- live updating using sidebar controls and time slider
- history manipulation - pushing/replacing history states

the most consequential changes are in cdc.js and map.js - most of the other stuff is basic markup adjustments


francisPLT:
I gave a quick look and also did some tests on demo2 instance. For now I don't have any comments.

I'll try and give a more thorough look, especially to get into the core logic, but I don't expect anything structural that would block you from moving forward.

PR #105, opened 2024-02-06
Clim 468 raster v1

perronld:
prepare_raster() implementations for v1.


francisPLT:
Looks good to me!

Only thing I would add is comments to indicate that `prepare_raster()` is implemented at different places and used by the Python API, so future developers don't break things accidently

francisPLT (approved):

PR #119, opened 2024-03-01
Clim 412 legend choro

perronld:
With this PR, I redesigned the legend rendering from scratch, and also used colours and values from the custom color scheme for legend and choropleth.


fw-child/resources/js/cdc.js
```diff
@@ -781,217 +783,123 @@
...
         let plugin = !this.item ? this.data('cdc_app') : this,
           options = plugin.options;
 
-        for (let i = options.legend.colormap.length - 1; i > 0; i--) {
-          if (d < options.legend.colormap[i].quantity) {
-            return options.legend.colormap[i].color;
+        let prev_colour = options.legend.colormap.colours[0],
+          prev_value = options.legend.colormap.quantities[0];
+
+        for (let i = 0; i < options.legend.colormap.colours.length; i++) {
```
perronld, line 888: I will rewrite this with a binary search... it will be around 4 times faster. Since get_color is called very often, it will make a difference.

phil-evans:
looks great!

PR #126, opened 2024-03-06
Clim 412 fix building climate zone

perronld:
I tried my best to make it as generic as possible for potential future variables.

### DIFF
--- a/fw-child/resources/js/map.js
+++ b/fw-child/resources/js/map.js
@@ -39,6 +39,7 @@
         colormap: {
           colours: [],
           quantities: [],
+          labels: null, // placeholder for special labels
         },
       },
       query: {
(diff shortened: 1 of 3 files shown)


fw-child/resources/js/map.js
```diff
@@ -1350,39 +1351,52 @@
...
       );
-      $.getJSON(
-        geoserver_url +
+
+      if (special_variables.hasOwnProperty(options.var_data.slug)) {
+        default_scheme_element.data('scheme-colours', special_variables[options.var_data.slug].colormap.colours);
+        default_scheme_element.data('scheme-quantities', special_variables[options.var_data.slug].colormap.quantities);
+        plugin.update_scheme();
+
+        if (typeof callback == 'function') {
```
xfrenette-crim, line 1363: Use 3 "=" to be consistent with other similar "if"
```suggestion
        if (typeof callback === 'function') {
```

fw-child/resources/js/map.js
```diff
@@ -1350,39 +1351,52 @@
...
-        options.query,
+      let default_scheme_element = item.find(
+        '#display-scheme-select .dropdown-item[data-scheme-id="default"]',
       );
-      $.getJSON(
-        geoserver_url +
+
+      if (special_variables.hasOwnProperty(options.var_data.slug)) {
+        default_scheme_element.data('scheme-colours', special_variables[options.var_data.slug].colormap.colours);
+        default_scheme_element.data('scheme-quantities', special_variables[options.var_data.slug].colormap.quantities);
```
xfrenette-crim, line 1360: Create a `special_var` variable, like in line 1409, just to lighten the first 2 lines of the if:
```suggestion
      if (special_variables.hasOwnProperty(options.var_data.slug)) {
        const special_var = special_variables[options.var_data.slug];
        default_scheme_element.data('scheme-colours', special_var.colormap.colours);
        default_scheme_element.data('scheme-quantities', special_var.colormap.quantities);
```

fw-child/resources/js/map.js
```diff
@@ -1350,39 +1351,52 @@
...
+        plugin.update_scheme();
+
+        if (typeof callback == 'function') {
+          callback();
+        }
+      } else {
+
+        // todo: implement behaviour for building_climate_zones
```
xfrenette-crim, line 1368: Is this TODO still relevant?

fw-child/resources/js/map.js
```diff
@@ -1350,39 +1351,52 @@
...
-            colour_map.map((e) => parseFloat(e.quantity)),
-          );
-          plugin.update_scheme();
-        })
-        .always(function () {
-          if (typeof callback == 'function') {
-            callback();
-          }
-        });
+
```
xfrenette-crim, line 1383: As per the _[boy scout rule](https://deviq.com/principles/boy-scout-rule)_, maybe we should remove the extra empty lines here.

xfrenette-crim (approved):
Approved once small changes implemented.

But the following may be desired for a future task:
* Even though the "Building Climate Zones" has a fixed palette, the UI still allows the user to change it
* If we select another palette, make it "Continuous" then come back to the default palette, the palette is back to "discrete", but the legend is still continous

fw-child/resources/js/map.js
perronld: I lost track of that line... is it good now?

PR #149, opened 2024-03-12
CLIM-468: Export map as PNG image

xfrenette-crim:
This PR implements the "Save map as image" button in the "Explore Maps" page.


perronld (approved):
Perfect and simple! I tested it in my dev environment and works flawlessly!

PR #154, opened 2024-03-13
CLIM-476 - Add a "Download" button in "Explore Maps"

xfrenette-crim:
In the _Download_ tab of the _Explore Maps_ page, it was decided to replace the _Export Data_ section with a single button that leads to the _Download Data_ page. This PR adds this button.

**Before**

**After**

<hr>

When clicking on the button, we want the user to be sent to the _Download Data_ page, but we also want this page to be prefilled with the values the user selected in this page. For example, we want the user's selected variable to "stay" selected.

### Implementation

- The _href_ of the button is the _Download Data_ URL, without any query string.
- The `query` object of the current page is transformed to a URL query string (using `$(document).cdc_app('query.obj_to_url', query)`)
- The JS builds a URL from the button's href and the built query string, and redirects to it.

### Note
The text above the button was written by me. It needs to be officially defined and translated. But it will be done in a future and separate PR.

### DIFF
--- a/fw-child/resources/js/map.js
+++ b/fw-child/resources/js/map.js
@@ -852,6 +852,18 @@
           .val($(this).attr('data-scheme-id'))
           .trigger('change');
       });
+      
+      // "Download" button
+
+      // Handle of the click on the "Download" button of the "Share" tab.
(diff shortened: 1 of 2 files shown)


phil-evans:
looks great!

PR #163, opened 2024-03-15
CLIM-477 bugfixes

perronld:


xfrenette-crim:
One of the bugs doesn't seem fixed (i.e. "_MSC Climate Normals : hide Colors options & legend_").

To reproduce:
1. In the "Explore Maps" page
2. Select a non-station variable (ex: _Minimum Temperature_)
3. Then change again the variable to select the _MSC Climate Normals_ variable or the _Rainfall IDF Data_

Expected:
* The legend disappears
* It's not possible to change the color palette

Actual:
* The legend is still visible, with the values of the previously selected variable
* It's possible to open the menu of the color palettes and change it (but it doesn't have any effect on the map).

fw-child/resources/js/map.js
```diff
@@ -560,19 +560,23 @@
...
-        .css('fill-opacity', options.legend.opacity);
+      if (this_pane == 'raster') {
+        options.legend.opacity = ui.value / 100;
+
+        // link legend's opacity
+        item
+          .find('.legend')
+          .find('.legendbox')
+          .css('fill-opacity', options.legend.opacity);
+      }
```
xfrenette-crim, line 579: My personal opinion is that changing the opacity of the data should not change the opacity of the legend. In my view, lowering the opacity of the data is not to change its color, it's to see what is below it, so, in that view, the legend doesn't need to be partly invisible.

xfrenette-crim (changes requested):
See the code comment.

Also, one of the bugs doesn't seem fixed (i.e. "_MSC Climate Normals : hide Colors options & legend_").

To reproduce:
1. In the "Explore Maps" page
2. Select a non-station variable (ex: _Minimum Temperature_)
3. Then change again the variable to select the _MSC Climate Normals_ variable or the _Rainfall IDF Data_

Expected:
* The legend disappears
* It's not possible to change the color palette

Actual:
* The legend is still visible, with the values of the previously selected variable
* It's possible to open the menu of the color palettes and change it (but it doesn't have any effect on the map).

perronld:
The legend and color scheme controls are now fixed.
But unfortunately, delta+aggregation control are now broken for building_climate_zones since commit 8a78bd29e689ce9a7e41adf49051b46dd35c6743

perronld:
@phil-evans I tried it, it won't work... set_controls() overrides it.

perronld:
After the previous merge, building_climate_zones works properly

fw-child/resources/js/map.js
xfrenette-crim: Could we do like in v1 and have an opaque white background behind the legend, so that when we lower its opacity, we don't see the map through the legend?

In v1:
image

In v2:
image

xfrenette-crim:
@perronld I don't know if related to the code in this PR, but when selecting "MSC Climate Normals 1981-2010" and then selecting another variable, like "Coldest day", it doesn't switch variable (the MSC Climate Normals are still shown).

xfrenette-crim (approved):
See last comment about a bug where we cannot change from MSC Climate Normals to another variable. If the bug is unrelated to this PR, it's approved, else it's approved once the bug is fixed.

PR #177, opened 2024-04-10
CLIM-525 fixed blanks in choro when value is higher than scheme highe…

perronld:
…st value

### DIFF
--- a/fw-child/resources/js/cdc.js
+++ b/fw-child/resources/js/cdc.js
@@ -1071,7 +1071,7 @@
         switch (i) {
           case -1:
             // fallback case in case style/legend is wrong
-            return 'rgba(0,0,0,0)';
+            return options.legend.colormap.colours[options.legend.colormap.colours.length-1];
           case 0:
             return options.legend.colormap.colours[0];
           default:


fw-child/resources/js/map.js
```diff
@@ -2389,6 +2389,16 @@
...
             low = high * -1;
+            // workaround to avoid legend with repeated values for very low range variables
+            if ((high - low) * 10**options.var_data.acf.decimals < scheme_length) {
+              low = -(scheme_length / 2.0);
+              high = scheme_length / 2.0;
+            }
+          } else {
+            // workaround to avoid legend with repeated values for very low range variables
+            if ((high - low) * 10**options.var_data.acf.decimals < scheme_length) {
+              high = low + scheme_length * 0**options.var_data.acf.decimals;
```
xfrenette-crim, line 2400: Shouldn't it be 10 instead of 0?
```suggestion
              high = low + scheme_length * 10**options.var_data.acf.decimals;
```

fw-child/resources/js/map.js
```diff
@@ -2389,6 +2389,16 @@
           if (selected_scheme_item.data('scheme-type') === 'divergent') {
             high = Math.max(Math.abs(low), Math.abs(high));
             low = high * -1;
+            // workaround to avoid legend with repeated values for very low range variables
```
xfrenette-crim, line 2392: There is also a `generate_ramp()` function in download.js. Should this code also be updated there?

xfrenette-crim (changes requested):

fw-child/resources/js/map.js
perronld: No, the download page doesn't have the custom color scheme option.

fw-child/resources/js/map.js
perronld: Good catch.. but I noticed a worse algebra  when doing the correction. My tests worked because there is currently no variables in this situation that has decimals > 0

xfrenette-crim (approved):

PR #211, opened 2024-05-07
CLIM-534 reversed legend vertically, default legend type and bugs wit…

perronld:
…h special_vars


fw-child/resources/js/cdc.js
```diff
@@ -1078,29 +1078,6 @@
                     url: choro_path,
                     dataType: 'json',
                     success: function (data) {
```
xfrenette-crim, line 1080: Is the `success` function still necessary since it's a kind of noop?

xfrenette-crim (approved):
Just check the 2 minor comments

fw-child/resources/js/cdc.js
perronld: Indeed... looks *promising*

PR #213, opened 2024-05-20
Update social logos in footer

sebastian-ew:
This adds a more recent version of the fontawesome library to allow the use of the X twitter icon.


xfrenette-crim (changes requested):
The code is good, no change to do there, but (I know it's surprising) the files in the `node_modules/` must also be committed. The reason is that there is no `npm install` done during the deployment process (files are copied as is).

Since the `node_modules/` is already in the .gitignore, you will need to do a forced push of the files. Try to commit only the minimum number of files. But be sure to test, because I see that it's not only the `all.min.css` file that needs to be committed, but also (at least) the `webfonts/` directory (maybe other files, to check).

**Note:**
After going live, our next main tasks concern improving the development process, and this includes using node packages as they must be used (no `node_modules/` in the Git, and `npm install` when deploying).

sebastian-ew:
I added the files from the node_modules folder that are needed.

xfrenette-crim (approved):

PR #214, opened 2024-05-20
News: Single page: better style post hero

sebastian-ew:
Changing single news hero styles.

### DIFF
--- a/fw-child/resources/scss/_vendor.scss
+++ b/fw-child/resources/scss/_vendor.scss
@@ -27,6 +27,7 @@
   flex-wrap: wrap;
   list-style: none;
   padding: 0;
+  margin-bottom: 0;
 
   li {
     display: flex;
--- a/fw-child/resources/scss/_vendor.scss
+++ b/fw-child/resources/scss/_vendor.scss
@@ -30,22 +30,29 @@
   margin-bottom: 0;
 
   li {
-    display: flex;
-    margin-bottom: 0.5rem;
-    min-width: 2.375rem;
-
-    &:not(.share-permalink-wrap) {
-      .label {
-        display: none;
-      }
+    &:not(:first-child):not(.share-permalink-wrap) {
+      margin-left: 0.75rem;
+    }
+    
+    &.share-permalink-wrap {
+      display: flex;
+      flex: 1;
+      margin-top: 0.15rem;
     }
 
-    &.share-permalink-wrap {
-      flex: 0 0 100%;
-      width: 100%;
-
+    a {
+      display: flex;
+      align-items: center;
+      padding: 4px 0;
+      
+      i {
+        margin-right: 0.25rem !important;
+        font-size: 0.85rem;
+      }
+      
       span {
-        font-size: 0.875rem;
+        font-size: 0.85rem;
+        line-height: 1;
       }
     }
   }
--- a/fw-child/resources/scss/_vendor.scss
+++ b/fw-child/resources/scss/_vendor.scss
@@ -30,14 +30,9 @@
   margin-bottom: 0;
 
   li {
-    &:not(:first-child):not(.share-permalink-wrap) {
-      margin-left: 0.75rem;
-    }
-    
-    &.share-permalink-wrap {
-      display: flex;
-      flex: 1;
-      margin-top: 0.15rem;
+    &:not(:last-child) {
+      margin-right: 0.75rem;
+      margin-bottom: 0.25rem;
     }
 
     a {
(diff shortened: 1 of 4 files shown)


fw-child/template/post-meta.php
```diff
@@ -1,57 +1,59 @@
-<div class="news-meta-item mb-3 mb-lg-4">
-	<h6 class="mb-3 all-caps text-blue-100"><?php _e ( 'Date', 'cdc' ); ?></h6>
-	
-	<?php 
-		
-		the_time ( __( 'F j, Y', 'cdc' ) );
-	
+<div class="news-meta-block p-3 border border-gray-600">
+	<div class="news-meta-item mb-3 mb-lg-4">
+		<h5 class="mb-3 all-caps text-blue-100"><?php _e ( 'Date', 'cdc' ); ?></h5>
```
MahieddineAk, line 3: @sebastian-ew I would not recommend not using headings here. You can instead use <span> for labels.

sebastian-ew: @MahieddineAk I made the change from heading tag to span. Thanks!

fw-child/resources/scss/_vendor.scss
```diff
@@ -27,23 +27,33 @@
...
+    &:not(:first-child):not(.share-permalink-wrap) {
+      margin-left: 0.75rem;
     }
-
+    
     &.share-permalink-wrap {
-      order: 1;
-      flex: 0 0 100%;
-      width: 100%;
+      display: flex;
```
xfrenette-crim, line 38: On my larger screen, the "Copy link" button is too close to the "LinkedIn" button and not vertically centered with the other buttons.

(Chrome Windows)

image

I see that the previous developer seems to have somewhat reused the CSS from version 1 (where this button had a different look) for version 2. But in version 2, it's not required that this button has a different style than the other social links, and not required that it be on its own line. So it could simply use the same styling as the other `<li>` of social links.

fw-child/template/post-meta.php
```diff
@@ -1,57 +1,59 @@
...
 	?>
-</div>
 
-<div class="news-meta-item mb-3 mb-lg-4">
-	<h6 class="mb-3 all-caps text-blue-100"><?php _e ( 'Author', 'cdc' ); ?></h6>
-	
+	<div class="news-meta-item mb-3 mb-lg-4">
+		<span class="mb-1 all-caps text-blue-100 d-block"><?php _e ( 'Topics', 'cdc' ); ?></span>
+
+		<p><?php
```
xfrenette-crim, line 33: The hover effect of those "Topics" links is not the same as the hover effect on the social links. It should be uniformized. Note: it can be done simply by changing this `<p>` with a `<div>` (just validate that the margins are still OK after change). 

(I know the hover effect is probably too subtle, but it can be made into its own task later on, since the same effect is applied in various parts of the site)

```suggestion
		<div><?php
```

fw-child/resources/scss/_vendor.scss
```diff
@@ -27,23 +27,33 @@
...
-    margin-bottom: 0.5rem;
-
-    &:not(.share-permalink-wrap) {
-      order: 2;
-
-      .label {
-        display: none;
-      }
+    &:not(:first-child):not(.share-permalink-wrap) {
+      margin-left: 0.75rem;
```
xfrenette-crim, line 34: I know that there will be updates in the Builder to add other responsive styling, and that the social links would maybe never end up being displayed one under the other, but maybe the `<li>` should have a _right_-margin instead of a _left_-margin, to prevent this possible misalignment if they fall on more than one line:

image

xfrenette-crim (changes requested):

fw-child/resources/scss/_vendor.scss
MahieddineAk: @xfrenette-crim we've unified the styles so the "Copy Link" will have same styling as other sharing buttons.

fw-child/template/post-meta.php
MahieddineAk: @xfrenette-crim this has been fixed, there is no need for `<div>` as well.

fw-child/resources/scss/_vendor.scss
MahieddineAk: @xfrenette-crim this has been fixed. We also added some bottom margin to not have condensed view.

xfrenette-crim (approved):
Approved. Please merge using a "Squash and merge" then delete the branch.

PR #217, opened 2024-05-28
#63733: Make footer sponsor logos responsive

MahieddineAk:


MahieddineAk:
@xfrenette-crim all your feedback have been addressed.

xfrenette-crim (approved):

PR #218, opened 2024-05-28
News: Single page: better style post hero

sebastian-ew:


fw-child/template/post-meta.php
```diff
@@ -1,5 +1,5 @@
-<div class="news-meta-block p-3 border border-gray-600">
-	<div class="news-meta-item mb-3 mb-lg-4">
+<div class="news-meta-block p-3 border border-gray-600 row">
+	<div class="news-meta-item mb-3 mb-lg-4 col-md-8 col-xl-16">
```
xfrenette-crim, line 2: There is no margin between the columns of a same line. Maybe once the builder is configured it won't be a problem (don't hesitate if it's the case), but for now, in my dev environment, the columns are glued together:

image

xfrenette-crim (changes requested):

fw-child/template/post-meta.php
sebastian-ew: @xfrenette-crim indeed, with the builder changes it won't be an issue, but in any case, I pushed a couple of code changes to prevent this issue.

xfrenette-crim (approved):

PR #221, opened 2024-06-07
#63902: Fix bug on Builder modal

MahieddineAk:
Fix bug on Builder modal when viewing custom field block.


MahieddineAk:
@xfrenette-crim a quick fix for a bug that prevent working properly on the builder. As it is a minor one, I will merge it, but feel free to review it and share your thoughts.

PR #224, opened 2024-06-14
#63110: glossary improvements

domcode-ew:
Add AZ menu and scrollspy
Sort french results by ASC

### DIFF
--- a/fw-child/resources/js/child-functions.js
+++ b/fw-child/resources/js/child-functions.js
@@ -191,10 +191,10 @@ const $ = jQuery;
 
     // SMOOTH SCROLL
 
-    let glossaryScrollOffset = 0;
+    let smoothScrollOffset = 0;
 
     if ($('#glossary-list-nav').length) {
-      glossaryScrollOffset = $('#glossary-list-nav').outerHeight();
+      smoothScrollOffset = $('#glossary-list-nav').outerHeight();
     }
 
     $('a.smooth-scroll').click(function(event) {
@@ -203,7 +203,7 @@ const $ = jQuery;
       const target = $($(this).attr('href'));
 
       if (target.length) {
-        const scrollPosition = target.offset().top - glossaryScrollOffset;
+        const scrollPosition = target.offset().top - smoothScrollOffset;
 
         $('html, body').animate({
           scrollTop: scrollPosition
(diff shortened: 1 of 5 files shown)


xfrenette-crim (changes requested):
See comments.

All of them don't need a re-review, except for the part where the code should consider "é" and "e" (and other accented letters) as the same.

So this is why this PR is "Request changes".

fw-child/template/glossary.php
```diff
@@ -14,13 +15,13 @@
 $glossary_query_args = array(
 	'post_type'      => 'definition',
 	'posts_per_page' => -1,
-	'order'          => 'asc'
+	'order'          => 'asc',
+	'orderby'		 => 'title'
```
xfrenette-crim, line 19: Tabs are used for spacing before the `=` instead of spaces (contrary to the other lines)

image

fw-child/resources/js/child-functions.js
```diff
@@ -189,6 +189,38 @@ const $ = jQuery;
       }
     }
 
+    // SMOOTH SCROLL
+
+    let glossaryScrollOffset = 0;
```
xfrenette-crim, line 194: Probably something like `smoothScrollOffset` could be more generalized, since the smooth scrolling could be used on other pages.

fw-child/resources/js/child-functions.js
```diff
@@ -191,10 +191,10 @@ const $ = jQuery;
 
     // SMOOTH SCROLL
 
-    let scrollOffset = 0;
+    let glossaryScrollOffset = 0;
 
     if ($('#glossary-list-nav').length) {
-      scrollOffset = $('#glossary-list-nav').outerHeight();
+      glossaryScrollOffset = $('#glossary-list-nav').outerHeight();
```
xfrenette-crim, line 197: No need to change because it's really not a big issue, but just note that the height of the nav is determined only once. If we resize the window and the nav becomes on two or three lines, it becomes taller, but the scroll offset won't be updated.

xfrenette-crim (approved):
See small comments.

fw-child/resources/js/child-functions.js
domcode-ew: You're right! We had no 2 or 3 lines nav at first when I coded this. I'll change it. Thanks :)

xfrenette-crim (approved):

PR #226, opened 2024-06-19
Dev 63848 gsap motion library

domcode-ew:
Have the old interactive post animations work with GSAP


fw-child/functions.php
```diff
@@ -100,6 +100,7 @@ function child_theme_enqueue() {
 	) {
 		wp_enqueue_style ( 'wp-block-library' );
 		wp_enqueue_style ( 'gutenberg' );
+		wp_enqueue_style ( 'global-styles' );
```
MahieddineAk, line 103: @domcode-ew it seems you can remove `is_singular( 'interactive' )` from the condition, since it is used via resource CPT.

domcode-ew: No it's not. It's actually used via interactive CPT.
This condition was already existing. I just placed the enqueue in it.

fw-child/resources/functions/post-types.php
```diff
@@ -205,7 +205,7 @@ function posttype_interactive() {
         'show_in_admin_bar'   => true,
         'show_in_nav_menus'   => true,
         'can_export'          => true,
-        'has_archive'         => true,
+        'has_archive'         => false,
         'exclude_from_search' => false,
         'publicly_queryable'  => true,
```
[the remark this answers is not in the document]
domcode-ew: @MahieddineAk Setting 'has_archive' to false does not make the post public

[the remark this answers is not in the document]
MahieddineAk: @domcode-ew I can see that the interactive posts are public and accessible. You can notice the "View" option when you hover over "interactive" posts in the WP admin table.

fw-child/functions.php
MahieddineAk: @domcode-ew there is no single page for interactive CPT, hence, no need to enqueue anything for that CPT. Removing that part off the condition will improve the old code.

fw-child/template/learn/interactive-content.php
```diff
@@ -4,6 +4,22 @@
...
 $interactive = get_field ( 'asset_post_' . $GLOBALS['fw']['current_lang_code'], get_the_ID() );
 
+// Enqueue GSAP and dependencies when id 'timeline' is present in content.
+// Essential for compatibility with old interactive resource post
+function detect_id_and_enqueue_script($content) {
+    $id_to_detect = 'id="timeline"';
+
+    if (strpos($content, $id_to_detect) !== false) {
+		wp_enqueue_script ( 'gsap' );
+		wp_enqueue_script ( 'scrolltrigger' );
```
xfrenette-crim, line 14: It seems that the _motion.page_ plugin already includes the GSAP and the ScrollTrigger scripts. Could we validate if we really need to re-include them here, else it loads two times each script from different URLs and with different versions.

Maybe we could instead enqueue the _motion.page_'s versions of the scripts (ex: `wp_enqueue_script ( 'mp-gsap-js' );`) ?

xfrenette-crim (approved):
Approved with comments.

fw-child/resources/js/scroll.js
```diff
@@ -0,0 +1,274 @@
...
+
+          if ($(id_pre + el.id).length) {
+
+            if (!$(id_pre + el.id).hasClass('entered')) {
+              $(id_pre + el.id).css('opacity', 0).addClass('entered');
+            }
+
+            if (el.effect == 'manual' ) {
+
+              this_to = {
```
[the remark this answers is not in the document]
domcode-ew: This is all legacy code that has been copied and pasted from the older theme to make the 4 pages work. It works but it not been thoroughly reviewed. Here, it seems like `this_to` is already declared on line 200 with a let so it doesn't have to be declared on line 210 and 230.

fw-child/resources/js/scroll.js
```diff
@@ -0,0 +1,274 @@
...
+
+                  }
+                }
+              };
+
+              tl.fromTo( id_pre + el.id, this_from, this_to, el.position );
+
+            } else {
+
+              this_to = {
```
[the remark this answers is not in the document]
domcode-ew: See upper comment. I can do the changes if required.

fw-child/functions.php
```diff
@@ -169,6 +169,12 @@ function child_theme_enqueue() {
 
 	wp_register_script ( 'select2', 'https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/js/select2.min.js', array ( 'jquery' ), null, true );
 
+	// GSAP
+
+	wp_enqueue_script ( 'gsap', 'https://cdnjs.cloudflare.com/ajax/libs/gsap/3.3.3/gsap.min.js', null, '3.3.3', true );
+	wp_enqueue_script ( 'scrolltrigger', 'https://assets.codepen.io/16327/ScrollTrigger.min.js?v=32', array ( 'gsap' ), null, true );
+	wp_enqueue_script ( 'scroll', $child_js_dir . 'scroll.js', array ( 'jquery','scrolltrigger' ), null, true );
```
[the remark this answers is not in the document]
domcode-ew: Oups! These here should be registering, not enqueuing. Good call! Also, they are meant to be enqueued not only in resource pages, but only resource pages that has old GSAP animation code copied/pasted. The condition is based on the presence of the element ID `timeline`. This logic happens in `interactive-content.php`. There is only 4 pages that need these scripts.

fw-child/template/learn/interactive-content.php
domcode-ew: With the modifications made to the functions.php about registering instead of enqueuing GSAP, they won't load at the same time because of the condition in the `interactive-content.php` file. I think it's a good idea to have GSAP loading independently from the plugin so these pages do not rely on this plugin to work properly. I can change this if you want me to. Let me know.

xfrenette-crim: Modifications accepted.

PR #239, opened 2024-07-10
CLIM-698: Add CookieYes script tag

xfrenette-crim:
## Context

The site will use _CookieYes_ for cookies acceptance. _CookieYes_ is added to the site by including a &lt;script> tag in the header. The script's URL contains an ID unique to the domain. Example: `https://cdn-cookieyes.com/client_data/<ID>/script.js`

Since we have a different domain for each language, we need to insert a different script tag for each language.

There exists a _CookieYes_ WordPress plugin, but it doesn't support different IDs for the same WordPress installation. So instead of using the plugin, we add the script tag programmatically.

## This PR

This PR contains the code to insert the script tag programmatically.

The ID used in the script's URL is defined in `default_config.php`. The script tag is then inserted in the head using a WordPress hook.

## Related ticket

https://ccdpwiki.atlassian.net/browse/CLIM-698

### DIFF
--- a/fw-child/default_config.php
+++ b/fw-child/default_config.php
@@ -12,6 +12,12 @@ function custom_global_vars()
     $vars['analytics_ua_fr'] = "G-CPGH39EQSH";
     $vars['googletag_id'] = "GTM-NJ7L4NR";
     $vars['ga_cross_domain'] = "";
+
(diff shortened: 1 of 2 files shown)


domcode-ew (approved):
All good! Ça marche bien et le script tag apparait bien en premier de tous, tel que requis par Cookie Yes

PR #247, opened 2024-07-24
CLIM-707: Fix Builder breaking page when deleting spacing rows

xfrenette-crim:
## Context

The Builder has a bug. When editing a block's settings, if we have a "Spacing" setting without any rows, and save the page, the page can no longer be rendered correctly.

To reproduce the bug:

* Connect to the WordPress admin
* “View” an already created page that can be edited with the Builder (ex: a News article)
* On any content (ex: a section, a container, a block, ...), click “Edit”
* Open the “Settings” tab
* If no “spacing” setting is already there:
  * Add a new “spacing” setting
  * Add some rows of spacing values
  * Save the field, save the page and reload
  * Re-open the settings tab of the content.
* For each spacing row, click the “Delete” button to delete the row, but don’t click the “X” to delete the “spacing” setting.
* Save the field, save the page and reload

Expected:
* The whole page shows up again, allowing to continue editing.

Actual:
* The page reloads, but only the blocks up to the block we edited above show up. All the following blocks don’t show up.
* Also, it’s not possible to edit the existing blocks.

## This PR

This PR solves the problem by removing the "Spacing" setting, when saving the block, if it doesn't contain any row.

When saving a block's settings, the JavaScript checks for "variation" settings, settings that are only made of a list of variations -- like "Spacing", and removes them if they don't have any variation (i.e. row).

**Also:**
* An unrelated but small fix is applied to line 1616, where 2 parenthesis were missing.

## Related ticket

https://ccdpwiki.atlassian.net/browse/CLIM-707

### DIFF
--- a/framework/resources/js/builder.js
+++ b/framework/resources/js/builder.js
@@ -1247,15 +1247,19 @@
 		/**
 		 * Remove, from a list fields, empty variation settings.
 		 *
-		 * Some settings are made only of a list of variations. When they have no variation, the setting is essentially
-		 * empty. This function removes those.
+		 * Some settings are made only of a list of variations. When they have
+		 * no variation, the setting is essentially empty. This function removes
+		 * those.
 		 *
-		 * @param {{name: String, value: String}[]} fields List of form fields containing setting fields.
-		 * @return {{name: String, value: String}[]} The list of form fields without the empty variation settings.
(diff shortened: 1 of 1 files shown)


domcode-ew (approved):
Working as expected. Good to go!

PR #254, opened 2024-07-31
CLIM-715: Fix: Builder strips page content when too many elements

xfrenette-crim:
## Context

There is a bug in the Builder. When saving a page that contains more than a certain number of elements (ex: sections, blocks, ...), the Builder "deletes" the last sections/elements, including the footer.

To reproduce:
* Create a new page, or edit an existing one.
* In the WordPress _Edit screen_ of the page, input the content of [this JSON file](https://github.com/user-attachments/files/16442880/builder.json) into the "Builder object" field.
* Save the page.
* "View" the page in the Builder.
* Without modifying anything, save the page using the Builder.
* Refresh the page.

Expected:
* The same content is shown, including the last sections and the footer.

Actual:
* The last sections of the page, including the footer were partly or completly removed.

## Reason of the bug

When saving the page, the Builder's JavaScript creates a POST request containing all the elements and their attributes. The POST request contains one parameter for each attribute of each element. When we have more than a certain number of elements, the number of parameters in the request can easily be over 1,000. For security reasons, PHP has a limit on the number of allowed parameters in a POST request. When it exceeds this number, PHP will simply strip the exceeding ones. The Builder's PHP code will thus receive only a limited list of elements and their attributes.

## This PR

This PR fixes the bug by changing how the the elements and their attributes are sent in the request. Instead of one parameter for each attribute, a single parameter is sent: the Builder object (containing the structure of the page) _stringified_ as a JSON object. The PHP then decodes this JSON object.

Also contains a small unrelated fix about missing parentheses.

## Related ticket

https://ccdpwiki.atlassian.net/browse/CLIM-715

### DIFF
--- a/framework/resources/functions/builder/ajax.php
+++ b/framework/resources/functions/builder/ajax.php
@@ -20,65 +20,61 @@ add_action ( 'wp_ajax_fw_output_loop_ajax', 'fw_output_loop_ajax' );
 // UPDATE POST
 //
 
+/**
+ * Update a post's content from a Builder JSON object in a POST request.
+ *
+ * From the POST request, retrieve the Builder JSON object in the `builder`
+ * argument and update the post's content accordingly.
+ */
 function fw_update_post() {
-	
-	// dumpit ( $_POST );
-	
 	$success = false;
-	
-	if (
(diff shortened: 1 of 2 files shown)


domcode-ew (approved):
Tested and working properly.

PR #255, opened 2024-07-31
CLIM-712: Fix: Apps are always shown only in English only

xfrenette-crim:
## Context

The v2 website doesn't allow a different URL for French for the embedded apps.

To reproduce:
* Visit the apps page in French (ex: https://donneesclimatiques2.crim.ca/apps/).
* Click on an app.

Expected:
* The app is shown in French.

Actual:
* The app is shown in English.

## This PR

This PR fixes the problem by adding an ACF field in the App post editor allowing to specify a French version of the URL. The code to embed the app was updated accordingly.

## Related ticket

https://ccdpwiki.atlassian.net/browse/CLIM-712

### DIFF
--- a/fw-child/acf-json/group_65efa3288aca3.json
+++ b/fw-child/acf-json/group_65efa3288aca3.json
@@ -203,11 +203,11 @@
             "endpoint": 0
         },
         {
-            "key": "field_65efa3289ea89",
-            "label": "App URL",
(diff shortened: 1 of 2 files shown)


domcode-ew (approved):
Works great! I might not have the right URL for FR, but I can see the FR url is being loaded in the iframe.

PR #278, opened 2024-08-26
CLIM-732: "Social links" Builder block: add option to use brand color…

xfrenette-crim:
## Context

A Builder block was recently created that allows to insert social links in a page. The links are shown as logos and their color is simply the default link color of where the block is inserted. So if the block is inserted in a page where links are red, we would see all the social icons red.

image

This can be desired in some places, but in other places, we may want to show the icons as using their "brand" color (blue for Facebook, black for "X", etc.)

## This PR

This PR adds a checkbox in the edit form of the "Social links" block allowing to show the icons with their "brand" color instead of the default link color.

image

## Related ticket

https://ccdpwiki.atlassian.net/browse/CLIM-732

### DIFF
--- a/fw-child/resources/functions/builder/front-end/block/content/social-links.php
+++ b/fw-child/resources/functions/builder/front-end/block/content/social-links.php
@@ -84,7 +84,9 @@ if ( isset( $element[ 'inputs' ][ 'links' ][ 'rows' ] ) ) {
 						target="_blank"
 						rel="noopener"
 						title="<?php echo esc_attr( $link[ 'label' ] ) ?>"
-						style="<?php echo ( ! empty( $link[ 'brand-color' ] ) ) ? "color: {$link[ 'brand-color' ]};" : "" ?>"
+						<?php if ( ! empty( $link[ 'brand-color' ] ) ) : ?>
+							style="color: <?php echo $link[ 'brand-color' ] ?>;"
+						<?php endif; ?>
 					>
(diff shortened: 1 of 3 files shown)


fw-child/resources/scss/_template.scss
```diff
@@ -442,4 +442,14 @@
...
+
+  &.brand-colored {
+    a {
+      transition: opacity 0.1s;
+    }
+
+    a:hover {
+      opacity: 0.6;
+    }
+  }
```
domcode-ew, line 454:
```suggestion

  &.brand-colored {
    a {
      transition: opacity 0.1s;
      
      &:hover {
        opacity: 0.6;
      }
    }
  }
```

fw-child/resources/functions/builder/front-end/block/content/social-links.php
```diff
@@ -71,6 +84,7 @@
 						target="_blank"
 						rel="noopener"
 						title="<?php echo esc_attr( $link[ 'label' ] ) ?>"
+						style="<?php echo ( ! empty( $link[ 'brand-color' ] ) ) ? "color: {$link[ 'brand-color' ]};" : "" ?>"
```
domcode-ew, line 87: I would put the whole style attr in the condition to avoid echoing an empty style attr.
I would remove the extra parentheses around the condition. (WCS: No Extra Parentheses)
I would remove the extra spaces inside the square brackets.

domcode-ew (approved):
Some minor syntax changes proposed for consistency. No second code review required.

PR #300, opened 2024-12-03
CLIM-783 Add a gitlab cicd pipeline to the repo

KevinPeymani-crim:
## Description

Add a gitlab cicd pipeline to the repo

## Related ticket

CLIM-783


xfrenette-crim (changes requested):

xfrenette-crim (changes requested):

.gitlab-ci.yml
```diff
@@ -39,63 +39,65 @@ build-push-portal-image:
...
-  - >
-    if [ ! -f ~/.ssh/known_hosts ]; then
-      touch ~/.ssh/known_hosts;
-      chmod 644 ~/.ssh/known_hosts;
-    fi
-  - export INSTANCE_HOST="$(ssh-keyscan -t rsa,ed25519 $INSTANCE_IP)"
   - >
     if ! grep -q $INSTANCE_HOST ~/.ssh/known_hosts; then
       $INSTANCE_HOST >> ~/.ssh/known_hosts;
     fi
```
xfrenette-crim, line 90: This command is not required anymore since we use we don't use the `known_hosts` file (because we use `StrictHostKeyChecking=no`)

.gitlab-ci.yml
```diff
@@ -0,0 +1,101 @@
...
+  - >
+    if [ ! -f ~/.ssh/known_hosts ]; then
+      touch ~/.ssh/known_hosts;
+      chmod 644 ~/.ssh/known_hosts;
+    fi
+  - export INSTANCE_HOST="$(ssh-keyscan -t rsa,ed25519 $INSTANCE_IP)"
+  - >
+    if ! grep -q $INSTANCE_HOST ~/.ssh/known_hosts; then
+      $INSTANCE_HOST >> ~/.ssh/known_hosts;
+    fi
```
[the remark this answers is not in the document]
xfrenette-crim: You should keep the `-o UserKnownHostsFile=/dev/null` as specified in the comment, because else the command will still add a new entry in the `known_hosts` file even if it doesn't use it.

.gitlab-ci.yml
```diff
@@ -0,0 +1,99 @@
...
+  before_script:
+  - command -v ssh-agent >/dev/null || ( apt-get update -y && apt-get install openssh-client -y )
+  - touch ~/.ssh/ccdp_deploy
+  - chmod 600 ~/.ssh/ccdp_deploy
+  - echo $SSH_PRIVATE_KEY | base64 --decode > ~/.ssh/ccdp_deploy
+
+  script:
+  - eval "$(ssh-agent)"
+  - ssh-add -k ~/.ssh/ccdp_deploy
+  - ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no $INSTANCE_USER@$INSTANCE_IP app-manager restart
```
xfrenette-crim, line 95: The key file can be passed directly to `ssh` without the need of using the `ssh-agent` (I think `-i`). If so, all lines related to the ssh-agent can be removed (the one that installs it, the ssh-agent line and the ssh-add line)

I would also take the opportunity to put the command on multiple lines:

```
- >
  ssh
  -o UserKnownHostsFile=/dev/null
  -o StrictHostKeyChecking=no
  -i <...>
  $INSTANCE_USER@$INSTANCE_IP
  app-manager restart
```

xfrenette-crim (changes requested):

xfrenette-crim (approved):

PR #308, opened 2025-01-23
refs #78879: fix dynamic import of leaflet.sync plugin for comparing …

andruxnet:
## Description

This PR will apply the following changes:
- It removes the need of dynamically importing the leaflet.sync plugin for comparing maps


mahieddineak (approved):
@andruxnet I will do tests on the generated data.

PR #309, opened 2025-01-26
refs #77333: create download app

andruxnet:
## Description

This PR contains the updated Download app, with the following working features:
- Step navigation via the "Next Step" button
- Step navigation via the Summary card
- Store to keep track of data selected in each step, and visible in the Summary card
- Map search control
- Map region drawing controls


mahieddineak (approved):
@andruxnet currently I will proceed with merging for the demo's sake today. I will review it afterwards and leave my feedback if any.

PR #312, opened 2025-01-29
#77333 Updated download app

andruxnet:
## Description

This PR contains the following features/fixes for the Download app.

- Correct styles and layout for all steps, including their logic to persist their data through the end of the form flow.
- Forward navigation logic, including a disabled "Next" button if the data in the current step is not filled in/valid.
- Backwards navigation logic via the Summary card, including resetting data to defaults for steps after step being navigated to.
- Field validation for numerical inputs where an empty value ("degrees" number input) or a `0` value ("decimal place" dropdown) is handled correctly, while before it was causing an error or just behaving inconsistently.


fw-child/apps/apps-src/src/components/download/step-send-request.tsx
mahieddineak, line 1: Are we missing in this file the send button?

fw-child/apps/apps-src/src/components/download/step-summary.tsx
```diff
@@ -4,54 +4,64 @@ import { PencilLine } from "lucide-react";
 
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
```
mahieddineak, line 5: This line is not translatable in this file (change not appear on this PR):
`<CardTitle className="text-xl font-semibold">Summary</CardTitle>`

fw-child/apps/apps-src/src/components/download/step-variable.tsx
```diff
@@ -32,21 +31,21 @@ const StepVariable: React.FC = () => {
...
-					label="Sectors"
-					tooltip="Select a sector"
-					placeholder="All"
-					value={filterValues.sector || ""}
+					label={__('Sectors')}
+					tooltip={__('Select a sector')}
+					placeholder={__('All')}
+					value={filterValues.sector || ''}
 				/>
 			</div>
```
mahieddineak, line 50: To make this similar to the map app, do you think we can add the search field as well?

fw-child/apps/apps-src/src/components/download/step-variable-options.tsx
```diff
@@ -7,17 +7,16 @@ import Dropdown from "@/components/ui/dropdown";
...
 const StepVariableOptions: React.FC = () => {
 	const { __ } = useI18n();
 
-	const dispatch = useAppDispatch();
-	const { version, degrees } = useAppSelector(state => state.download);
+	const { setField, fields } = useDownloadContext();
+	const { version, degrees } = fields;
 
 	const options = [
 		'CMIP5',
```
mahieddineak, line 22: On this file, this is not translatable:

`Please set your variables options to your needs.`

fw-child/apps/apps-src/src/context/download-provider.tsx
```diff
@@ -2,31 +2,106 @@
...
+
+			// we need some custom validation for some values
+
+			// email should be valid
+			if (key === 'email') {
+				return isValidEmail(String(value));
+			}
+
+			// a zero is valid for number types, except in step 4 where selectedCells are required
+			if (typeof value === 'number' && key !== 'selectedCells') {
```
mahieddineak, line 78: This doesn't look like a robust solution; I would like to have a dedicated validation for each input. Nothing to do now, but keep it in mind when working with live data.

fw-child/apps/apps-src/src/lib/utils.ts
```diff
@@ -4,3 +4,8 @@ import { twMerge } from 'tailwind-merge';
 export function cn(...inputs: ClassValue[]) {
 	return twMerge(clsx(inputs));
 }
+
+export const isValidEmail = (email: string): boolean => {
+	const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]{2,}$/;
```
mahieddineak, line 9: I am not recommending using this solution, but instead using a dedicated library such as: https://www.npmjs.com/package/validator

mahieddineak (changes requested):
@andruxnet LGTM, with some comments. Also could you please resolve merge conflicts?

fw-child/apps/apps-src/src/components/download/step-send-request.tsx
andruxnet: @mahieddineak we're not. The button is in `app.tsx`.

fw-child/apps/apps-src/src/components/download/step-variable.tsx
andruxnet: @mahieddineak the search input in these dropdowns was disabled also from the map app, because after some of the changes to fix the build errors it started working very inconsistently -- losing focus after each keystroke I think was the main issue.

Once fixed in the map app we can make these searchable as well.

mahieddineak: @andruxnet I am referring to this search field:
Cursor_and_CLIM_-_High_Fidelity_Mockups_UX_UI_–_Figma

andruxnet: @mahieddineak I did not see that search field when I initially built this panel, so it's not implemented at all yet... maybe we need to create a new task for this feature?

fw-child/apps/apps-src/src/lib/utils.ts
andruxnet: @mahieddineak I installed the `validato`r package as suggested, and changed the `emailRegex` solution for the `validator.isEmail` way.

andruxnet:
@mahieddineak All conflicts have been fixed, and your feedback has been addressed.

fw-child/apps/apps-src/src/context/download-provider.tsx
mahieddineak: @andruxnet this doesn't seem to be addressed. I also find in this file that isValidEmail is used instead of the dedicated function by the third-party library.

mahieddineak:
@andruxnet I'm going to merge this PR, but as you are still working on this ticket, please see my last comment on the `context/download-provider.tsx` file.

fw-child/apps/apps-src/src/context/download-provider.tsx
andruxnet: @mahieddineak that line wasn't changed because you said

> Nothing to do now, but keep it in mind when working with live data.

Also, the email validation you see here is in the old location under `apps/apps-src`, the new location already uses the `validator` package instead as you can see here: https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/blob/dev-78677-apps/apps/src/lib/utils.ts

PR #319, opened 2025-02-10
refs #76256: Initialize apps JSON structures

mahieddineak:
## Description

This PR contains the initial JSON structure for configs needed for the map and download app to work dynamically. It contains:
* ClimateData variable configuration.
* The map and download global configs, like API endpoints.
* WordPress REST API responses:
  * List of datasets.
  * List of variables.
  * Single variable post.

## Related ticket
* https://rm.ewdev.ca/issues/76256

## Note
* This PR isn't intended for merging.

### DIFF
--- a/JSON/apps/variables/var-UNIQUE_GLOBAL_VAR_ID-1.json
+++ b/JSON/apps/variables/var-UNIQUE_GLOBAL_VAR_ID-1.json
@@ -19,6 +19,7 @@
       "enabled": true,
       "standard": true,
       "station": true,
+      "grid": "canadagrid",
       "props": {
         "standard": [
           "sea_level"
--- a/JSON/apps/variables/var-UNIQUE_GLOBAL_VAR_ID-1.json
+++ b/JSON/apps/variables/var-UNIQUE_GLOBAL_VAR_ID-1.json
@@ -1,0 +1,1205 @@
+{
+  "id": "UNIQUE_GLOBAL_VAR_ID_SET_ON_WP",
+  "props": {
+    "ids": {
+      "wp": {
+        "post_id": 123
+      },
+      "map": {
+        "standard": "MAP_STANDARD_ID",
+        "station": "MAP_STATION_ID"
+      },
+      "download": {
+        "precalculated": "DOWNLOAD_PRECALCULATED_ID",
+        "analyze": "DOWNLOAD_ANALYZE_ID",
+        "station": "DOWNLOAD_STATION_ID"
+      }
+    },
+    "map": {
+      "enabled": true,
+      "standard": true,
+      "station": true,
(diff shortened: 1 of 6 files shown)


JSON/apps/variables/var-UNIQUE_GLOBAL_VAR_ID-1.json
```diff
@@ -0,0 +1,1206 @@
+{
+  "id": "UNIQUE_GLOBAL_VAR_ID_SET_ON_WP",
```
xfrenette-crim, line 2: Multiple "precalculated" variables are _variations_ of the same variable, with a different threshold. For example, "Days above Tmax" have the following values for Tmax: 25, 27, 29, 30, 32.

In the backend, they are considered as 5 different variables (`txgt_25`, `txgt_27`, `txgt_29`, ...), but in the UI, we want to show only a single variable ("Days above Tmax") and show a dropdown to select the threshold:

image

image

I don't see how this would be represented in this file.

JSON/apps/variables/var-UNIQUE_GLOBAL_VAR_ID-1.json
xfrenette-crim, line 1: For some fields and attributes, their value is not "static", it depends on some logical condition calculated on the current UI state.

Examples:
* The list of scenarios of a variable depends on the "dataset version" currently selected (CMIP6 or CMIP5).
* For the Relative Sea Level Change variable, the `period` values depends on the selected scenario.
* For some "precalculated" variables, if the user selects the "Daily" frequency, it then becomes an "analysis" variable.

New variables will likely require new kinds of logic, still unkown.

The problem I see here is that we are trying to create a JSON structure that encodes those logics, and then require the JavaScript to parse this JSON and correctly find the correct value based on user interaction.

This seems to create a complex JSON, and I guess it may require a complex JavaScript parser.

At the same time, it seems that those logics could be expressed simply in JavaScript functions, executed at the time of generating the values. 

For example: for a certain variable, a specific JavaScript function is designated as the one returning the list of scenarios. This function has access to the currently selected fields (i.e. the selected version), so it can easily do some simple logic to return the correct set of scenarios. 

The whole complexity of the logic for the scenario field would now all be contained inside a relatively simple function (few simple to understand lines, easier to maintain, to change, to test). And the code calling this function doesn't have to manage any logic.

New variables may come up with new kinds of logic (ex: the value of this feature cannot be above a certain value if the scenario is XYZ), and I'm afraid that trying to represent all of them in a custom JSON structure will make this file and its structure more complex every time.

mahieddineak:
I am closing this PR, as we've moved to a new hybrid solution. @don-ew, @xfrenette-crim.

PR #322, opened 2025-02-11
refs #79426 Use correct search API for Search Control component

andruxnet:
## Description

This PR will add the correct API location search endpoint for the Search Control component.

### DIFF
--- a/apps/src/components/map-layers/search-control.tsx
+++ b/apps/src/components/map-layers/search-control.tsx
@@ -3,10 +3,11 @@
  * This component allows users to search for locations using the OpenStreetMap Nominatim API and navigate the map to the selected location.
  */
 
-import { useState, useEffect, ReactElement, useCallback } from 'react';
+import { useState, useEffect, ReactElement, useCallback, useMemo } from 'react';
 import { useI18n } from '@wordpress/react-i18n';
 import { Locate, LocateFixed } from 'lucide-react';
 import { useMap } from 'react-leaflet';
+import { nanoid } from 'nanoid';
 
 import L from 'leaflet';
 import 'leaflet-search/dist/leaflet-search.min.css';
@@ -23,7 +24,7 @@ import {
 import {
 	SEARCH_PLACEHOLDER,
 	SEARCH_DEFAULT_ZOOM,
-	MAP_SEARCH_URL,
+	LOCATION_SEARCH_ENDPOINT,
 } from '@/lib/constants.ts';
 
 /**
@@ -49,7 +50,7 @@ export default function SearchControl({
 
 	// we need a unique id for the search control container for cases where multiple maps
 	// are rendered on the same page -- ie. comparing emission scenarios
-	const searchControlId = Math.random().toString(36).substring(2, 10);
+	const searchControlId = useMemo(() => nanoid(), []);
 
 	// defining default placeholder here so that it can be translated
 	const textPlaceholder = __(SEARCH_PLACEHOLDER) || '';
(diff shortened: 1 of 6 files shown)


apps/src/components/map-layers/search-control.tsx
```diff
@@ -114,19 +118,56 @@ export default function SearchControl({
 			return;
 		}
 
+		const buildLocationTitle = (item: SearchControlLocationItem) => {
```
mahieddineak, line 121: Is this logic similar to V2? Could you please document function so we can see an example at glance.

apps/src/lib/constants.ts
```diff
@@ -33,6 +33,5 @@ export const SLIDER_DEFAULT_YEAR_VALUE: number = 1971;
 export const SEARCH_PLACEHOLDER: string =
 	'Zoom to a location, region, city, coordinates...';
 export const SEARCH_DEFAULT_ZOOM: number = 10;
-export const SEARCH_COUNTRY_CODE: string = 'ca';
-export const MAP_SEARCH_URL: string = `https://nominatim.openstreetmap.org/search?format=json&q={s}&countrycodes=${SEARCH_COUNTRY_CODE}`;
+export const MAP_SEARCH_URL: string = '/wp-json/cdc/v2/location_search/?q={s}';
```
mahieddineak, line 36: As this is not only for map, it is better to name it LOCATION_SEARCH_ENDPOINT

apps/src/components/map-layers/search-control.tsx
```diff
@@ -45,7 +49,7 @@ export default function SearchControl({
 
 	// we need a unique id for the search control container for cases where multiple maps
 	// are rendered on the same page -- ie. comparing emission scenarios
-	const searchControlId = Math.random().toString(36);
+	const searchControlId = Math.random().toString(36).substring(2, 10);
```
mahieddineak, line 52: I think it is better if we use dedicated library to generate random strings, such us `nanoid`

mahieddineak (changes requested):
@andruxnet this PR looks good, however, it contains bugs on the UI:
* It is not working consistently, for example, searching for ottawa triggers console errors (see below).
* Suggestions aren't wrapping to the next line and get cropped.
* Highchart appears on the download app, which is not expected. Also, no need for tooltip of median and range.
* I left you some comments on the PR as well.

»_Download_App

apps/src/components/map-layers/search-control.tsx
andruxnet: @mahieddineak Yes, I used the same logic from V2, where the location title is composed of different terms from the location search result, in the form of "text (term) location, province".

Examples:
Ottawa (City) Carleton; Russel, Ontario
Nottawasaga River (River), Simcoe, Ontario

image
image

andruxnet:
@mahieddineak feedback addressed. Also, regarding this particular feedback:

> @andruxnet this PR looks good, however, it contains bugs on the UI:
> 
> * It is not working consistently, for example, searching for ottawa triggers console errors (see below).
> * Suggestions aren't wrapping to the next line and get cropped.
> * Highchart appears on the download app, which is not expected. Also, no need for tooltip of median and range.
> * I left you some comments on the PR as well.
> 
> »_Download_App

The issue with the console errors comes from the `cdc_location_search` function in `fw-child/resources/functions/rest.php` that is not `die()`ing after echoing the JSON response to the request. This causes a "headers already sent" error and breaks the JSON response. I explained this in [this Slack thread](https://evolvingweb.slack.com/archives/C07ME79QCT1/p1739303118709499).

Feedback address from that top feedback:
> * Suggestions aren't wrapping to the next line and get cropped. ===> DONE
> * Highchart appears on the download app, which is not expected. Also, no need for tooltip of median and range. ===> DONE

PR #324, opened 2025-02-17
refs 80333: Implement cells and region selection modes to download app

andruxnet:
## Description

This PR will implement both **Grid Cells** and **Draw Region** selection modes to the Download app Location step.
It also adds a **Clear** button to clear selections, and it also persists selections when going back to the Location step from the Summary card.


apps/src/components/download/step-location.tsx
```diff
@@ -85,58 +104,66 @@ const StepLocation: React.FC = () => {
...
 
 			<MapContainer
-				center={CANADA_CENTER}
+				center={center}
 				zoomControl={false}
-				zoom={DEFAULT_ZOOM}
+				zoom={zoom}
 				minZoom={DEFAULT_MIN_ZOOM}
 				maxZoom={DEFAULT_MAX_ZOOM}
 				scrollWheelZoom={false}
```
mahieddineak, line 158: Is there a reason to have this as false? It is a nice feature.

apps/src/components/download/step-location.tsx
```diff
@@ -54,6 +57,22 @@ const StepLocation: React.FC = () => {
 		{ value: 'watershed', label: __('Watersheds') },
 	];
 
+	const clearSelection = () => {
+		gridLayerRef.current?.clearSelection();
+	};
+
+	const renderGrid = () => {
```
mahieddineak, line 64: This logic seems missing a couple of things:
* The Grid cells and Draw region options should appear only when the interactive regions are selected.
* Depending on other interactive regions (other than Grid Cells), the selection on the map changes, please see here: https://climatedata.ca/analyze/
* The count of cells should change depending on the interactive regions selected, so it could be like:
  * 23 Cells
  * 2 watersheds
  * 8 census subdivisions
  * 3 health regions
* Regarding the count for regions other than the grid option, it seems it is always one region that could be selected, and it is not possible to select more than one. I asked the client, and I will update this comment once I receive an answer.

Cursor_and_Analyze_Climate_Data_—_ClimateData_ca

Tiling_Handle_Window_and_Analyze_Climate_Data_—_ClimateData_ca

Analyze_Climate_Data_—_ClimateData_ca

apps/src/components/download/step-location.tsx
```diff
@@ -54,6 +57,22 @@ const StepLocation: React.FC = () => {
...
+		if (interactiveRegion === 'gridded_data') {
```
mahieddineak, line 65: Instead of using strings, I suggest creating 4 constants that define all the 4 possible regions.

apps/src/components/download/step-location.tsx
```diff
@@ -85,58 +104,66 @@ const StepLocation: React.FC = () => {
...
+								<div
+									className={cn(
+										'text-2xl font-semibold leading-7',
+										selectedCellsCount > 0
+											? 'text-brand-blue'
+											: 'text-neutral-grey-medium'
+									)}
+								>
+									{_n(
+										'1 Cell',
```
mahieddineak, line 141: Please refer to my comment above; it isn't only one cell.

apps/src/components/download/step-summary.tsx
```diff
@@ -17,7 +17,7 @@ const StepSummary: React.FC = () => {
 		variable,
 		version,
 		degrees,
-		selectedCells,
+		selectedCellsCount,
```
mahieddineak, line 20: This constant should have a generic name, let's say, selectedAreasCount.

apps/src/components/download/step-variable.tsx
```diff
@@ -64,7 +65,7 @@ const StepVariable: React.FC = () => {
 					filterValues={filterValues}
 					selected={variable}
 					onSelect={(selected) => {
-						setField('variable', selected);
+						dispatch(setVariable(selected));
```
mahieddineak, line 68: Same comment as above for the "learn more" link in the cards (also please fix that on the map app as well).

apps/src/components/download/step-variable.tsx
mahieddineak, line 1: We are still missing the search field here; I think we planned to do that later.

apps/src/components/download/step-variable-options.tsx
```diff
@@ -9,16 +9,17 @@ import Dropdown from '@/components/ui/dropdown';
...
  */
 const StepVariableOptions: React.FC = () => {
 	const { __ } = useI18n();
 
-	const { setField, fields } = useDownload();
-	const { version, degrees } = fields;
+	const { version, degrees } = useAppSelector((state) => state.download);
+	const dispatch = useAppDispatch();
 
 	const options = ['CMIP5', 'CMIP6'];
```
mahieddineak, line 24: These options aren't static and will be provided by the variable JSON config; also, they have a label and a value.

apps/src/components/download/step-variable-options.tsx
```diff
@@ -62,12 +63,13 @@ const StepVariableOptions: React.FC = () => {
 					<Input
```
mahieddineak, line 63: Does this field have + and -? And does it have a step attribute?

apps/src/components/map-layers/cells-grid-layer.tsx
```diff
@@ -100,29 +104,11 @@ const CellsGridLayer: React.FC = () => {
...
-
-			const datasetKey = dataset as keyof typeof scenarioNames;
+			const datasetKey = dataset as keyof typeof SCENARIO_NAMES;
 			const emissionKey =
-				emissionScenario as keyof (typeof scenarioNames)[keyof typeof scenarioNames];
+				emissionScenario as keyof (typeof SCENARIO_NAMES)[keyof typeof SCENARIO_NAMES];
 
-			const rcp = scenarioNames[datasetKey][emissionKey]
+			const rcp = SCENARIO_NAMES[datasetKey][emissionKey]
 				.replace(/[\W_]+/g, '')
```
mahieddineak, line 112: This looks fragile and is unnecessary as it will be provided by the variable JSON config.

apps/src/components/map-layers/selectable-rectangle-grid-layer.tsx
```diff
@@ -0,0 +1,168 @@
...
+		(e: L.LeafletEvent & { layer: L.Layer }) => {
+			// clear previous shapes
+			layerGroupRef.current.clearLayers();
+			layerGroupRef.current.addLayer(e.layer);
+
+			// calculate initial cell count
+			updateCellCount(e.layer);
+
+			// enable resizing
+			e.layer.pm.enable({
```
mahieddineak, line 91: This seems not working on UI.

apps/src/components/map-layers/selectable-rectangle-grid-layer.tsx
```diff
@@ -0,0 +1,168 @@
...
+				allowSelfIntersection: false,
+				draggable: true,
+			});
+
+			// disable drawing mode
+			map.pm.disableDraw('Rectangle');
+
+			// listen for rectangle edits to recalculate
+			e.layer.on(
```
mahieddineak, line 100: This doesn't seem to be working on UI.

apps/src/components/map-layers/selectable-rectangle-grid-layer.tsx
mahieddineak, line 1: Add file doc.

apps/src/components/map-layers/selectable-cells-grid-layer.tsx
mahieddineak, line 1: Add file doc.

apps/src/context/download-provider.tsx
```diff
@@ -54,9 +71,67 @@ export const DownloadProvider: React.FC<{ children: React.ReactNode }> = ({
 	const dispatch = useAppDispatch();
 	const state = useAppSelector((state) => state.download);
 
+	// helper method used to reset step fields when moving back to a previous step
```
mahieddineak, line 74: As I mentioned in one of my previous comments, this should be dynamic, as not all variables have the same fields.

apps/src/features/download/download-slice.ts
```diff
@@ -32,18 +38,88 @@ const downloadSlice = createSlice({
 	name: 'download',
 	initialState,
 	reducers: {
```
mahieddineak, line 40: Same comment as above; the reducers should be dynamic.

apps/src/lib/constants.ts
```diff
@@ -36,3 +36,113 @@ export const SEARCH_DEFAULT_ZOOM: number = 10;
 export const SEARCH_COUNTRY_CODE: string = 'ca';
 export const MAP_SEARCH_URL: string = `https://nominatim.openstreetmap.org/search?format=json&q={s}&countrycodes=${SEARCH_COUNTRY_CODE}`;
 export const OWS_FORMAT: string = 'image/png';
+
+export const SCENARIO_NAMES = {
```
mahieddineak, line 40: Scenario and datasets should be dynamic, but it is fine to keep them as they are now. 

Q: The low, medium and high values, could you please refer when you use them in the app? Or did you just copy them from V2?

mahieddineak (commented):
@andruxnet sounds good, thanks! I left you a couple of comments.

apps/src/components/download/step-location.tsx
mahieddineak: @andruxnet the client has confirmed that for regions other than grids, we can only select one area; that said, please remove numbers in those cases.

apps/src/components/download/step-dataset.tsx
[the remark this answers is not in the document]
andruxnet: @mahieddineak we can remove the RadioCardFooter altogether then, just please confirm because I do see that link in the Figma file.. unless you mean, that when there's no link for the dataset or variable then hide the footer, but show it when there is a link?

image

apps/src/components/download/step-location.tsx
andruxnet: @mahieddineak no reason, this was not intended so I'll set it back to true

apps/src/components/download/step-variable.tsx
andruxnet: Need confirmation to [this comment](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/324#discussion_r1960866486)

apps/src/components/download/step-variable.tsx
andruxnet: Yes, the search field is part of another ticket

apps/src/components/download/step-variable-options.tsx
andruxnet: It's a "regular" number input, so it does have -/+ buttons. No step attribute so it defaults to 1 as step.

image

apps/src/components/map-layers/cells-grid-layer.tsx
andruxnet: Will fix once the variable JSON config is made available.

apps/src/context/download-provider.tsx
andruxnet: Will fix this once we're able to pull data from the API

apps/src/features/download/download-slice.ts
andruxnet: Will fix this once we're able to pull data from the API

apps/src/lib/constants.ts
andruxnet: Those were copied over from v2.

andruxnet:
@mahieddineak Feedback has been addressed. I added a couple of questions on feedback where I wasn't sure how to proceed. Some feedback was left for when the app is able to pull data from the API to prevent duplicating efforts.

apps/src/components/download/step-dataset.tsx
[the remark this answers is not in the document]
mahieddineak: @andruxnet yes, I think it is fine if we remove all the card footer if no link has been introduced.

PR #327, opened 2025-02-21
refs #73776: Add missing features and correct styling for location info chart

andruxnet:
## Description

This PR will add missing features like navigator area and series toggle points at the bottom of the chart. It also adds correct styling to match the design.


mahieddineak:
@andruxnet LGTM. I note that there are still some items missing on the highchart; we will discuss them internally.

PR #328, opened 2025-02-23
refs #79405: Adjust taxonomy fetching logic to use WP REST API instead of dummy json file when available

andruxnet:
## Description

This PR will have the Datasets card pull data from the WP REST API instead of the dummy JSON file.


mahieddineak:
@andruxnet all good! I will merge the PR for now, and I've left you a minor comment to address in a separate PR.

PR #332, opened 2025-02-28
CLIM-882 : Use download script for docker assets

SadiaD-Crim:
## Description
This pull request uses the script that automates the download of SSL files and WordPress plugins for Docker assets, replacing the direct plugin download in the pipeline configuration.

## Related ticket
CLIM-882


xfrenette-crim (approved):
Approuvé, mais ne pas oublier d'enlever le `if` utilisé uniquement pendant le développement.

PR #344, opened 2025-03-16
refs #81983: Add missing interactive region layers to the map

andruxnet:
## Description

This PR will add missing interactive region layers to the map app.


apps/src/components/map-layers/interactive-regions-layer.tsx
```diff
@@ -0,0 +1,279 @@
...
+		variable,
+		dataset,
+		decade,
+		frequency,
+		interactiveRegion,
+		emissionScenario,
+		legendData,
+	} = useAppSelector((state) => state.map);
+
+	// TODO: revisit this because colors are not correct.. need to check the original implementation
```
mahieddineak, line 46: Does this todo still relevant?

apps/src/components/map-layers/interactive-regions-layer.tsx
```diff
@@ -0,0 +1,279 @@
+import React, {
```
mahieddineak, line 1: Please add file description, mentioning all the interactive region types.

apps/src/components/map-layers/interactive-regions-layer.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+		return {
+			colours: colormapEntries.map((entry) => entry.color),
+			quantities: colormapEntries.map((entry) => Number(entry.quantity)),
+			scheme_type: 'continuous',
+		};
+	}, [legendData]);
+
+	// Function to interpolate between colors
+	const interpolate = useCallback(
```
mahieddineak, line 64: Please specify all functions that are legacy code, and if possible the file path and line of code.

apps/src/hooks/use-interactive-map-events.tsx
```diff
@@ -0,0 +1,343 @@
+import { useMemo, useRef } from 'react';
```
mahieddineak, line 1: Please add file documentation and mention this is a legacy code.

apps/src/lib/constants.ts
```diff
@@ -43,7 +43,8 @@ export const REGION_CENSUS: string = 'census';
 export const REGION_HEALTH: string = 'health';
 export const REGION_WATERSHED: string = 'watershed';
 
-// TODO: these will come from the API
+// TODO: these will come from the API.. also, "very-high" key was added because when choosing
```
mahieddineak, line 46: I've met with very high before, which in your code seems the same as high, could you please point out where you find this in V1 and V2?

fw-child/resources/js/cdc.js
```diff
@@ -1392,6 +1392,8 @@
                               style_obj.fillOpacity = 1;
                             }
 
+                            console.log('aaaaaaa',style_obj)
```
mahieddineak, line 1395: Please remove this line.

fw-child/resources/js/map.js
```diff
@@ -957,6 +957,7 @@
 
           options.grid.highlighted = shape_id;
 
+          console.log('aaaa', shape_id, style_obj);
```
mahieddineak, line 960: Please remove this line.

mahieddineak (commented):
@andruxnet thanks! a couple of comments left on the PR.

apps/src/components/map-layers/interactive-regions-layer.tsx
andruxnet: no, I forgot to remove it

apps/src/lib/constants.ts
andruxnet: The "very-high" option comes from the Figma designs, but the current implementation uses only low, medium, and high. The value for "very-high" in the Figma designs is actually the value used for "high" in the current implementation.

I've removed "very-high" option now, because following the rules for building the request based on this value didn't seem to work, so I left this with only the working options according to the original implementation.

mahieddineak: @andruxnet thanks for explaining. In reality, there are no very-high emissions, only: low, medium, and high.

PR #346, opened 2025-03-19
refs #82948: now able to use a custom domain to query WP for dataset.

PierrePaul:
## Description

This add the ability to have a custom domain for the WordPress API. It will default to https://dev-en.climatedata.ca which is currently a loopback address.

## Related ticket

https://rm.ewdev.ca/issues/82948 - Set up staging for CLIM to test the code

### DIFF
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -62,8 +62,6 @@ services:
     image: climatedata-task-runner
     build:
       context: .
-      args:
-        VITE_WP_API_DOMAIN: https://dev-en.climatedata.ca
       dockerfile: dockerfiles/build/task-runner/Dockerfile
     command: ["bash", "-c", "watch-fe.sh /app"]
     volumes:
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -62,6 +62,8 @@ services:
     image: climatedata-task-runner
     build:
       context: .
+      args:
+        VITE_WP_API_DOMAIN: https://dev-en.climatedata.ca
       dockerfile: dockerfiles/build/task-runner/Dockerfile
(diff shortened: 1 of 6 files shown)


docker-compose-dev.yaml
```diff
@@ -62,6 +62,8 @@ services:
     image: climatedata-task-runner
     build:
       context: .
+      args:
```
mahieddineak, line 65: @PierrePaul, as I understand from your PR, it looks like our staging server isn't loading datasets and variables (and probably other resources) properly because the WP home URL is hardcoded. While your solution is proper, it needs to update the _docker-compose-dev.yaml_ on each environment separately (usually we don't change the _docker-compose-dev.yaml_ as it is tracked by git; so we use a _docker-override.yaml_ file).

As we have a couple of different environments, it seems harder to maintain them. The simplest solution within the project context is to pass the WP home URL similar to how we did with the language to the React root: https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/blob/c01ab1eea336603a46a0e76774195149770e9045/fw-child/apps/app-map.php#L63

Example:
```
<div id="root"   
     data-app-lang="<?php echo esc_attr( $current_lang ); ?>"   
     data-wp-home-url="<?php echo esc_attr( WP_HOME ); ?>">  
</div>
```
For that, there is no need to change any Docker or environment related files.

mahieddineak (changes requested):
@PierrePaul I left you a comment on the PR, thanks!

docker-compose-dev.yaml
PierrePaul: @mahieddineak PR updated! Thanks for the review

mahieddineak (commented):
@PierrePaul Great! Just for consistency, we already have a dedicated file for configs, on _src/lib/constants.ts_

PR #348, opened 2025-03-24
refs #78677: Pass the WP post ID.

don-ew:
## Description

Pass the post ID from the WP API to local variable data.

### DIFF
--- a/apps/src/types/types.ts
+++ b/apps/src/types/types.ts
@@ -121,7 +121,7 @@ export type ApiPostData = {
  */
 export type PostData = {
 	id: string | number;
-	wpId: number;
+	postId: number;
 	title: string;
 	description?: string;
 	link?: { title: string; url: string; target: string };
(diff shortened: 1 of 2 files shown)


apps/src/types/types.ts
```diff
@@ -120,6 +121,7 @@ export type ApiPostData = {
  */
 export type PostData = {
 	id: string | number;
+	wpId: number;
```
mahieddineak, line 124: For consistency, could we let all of them be `post_id`? Which refers to WordPress post id, or further `wp_post_id`. I can change the REST API response based on that.

mahieddineak (commented):
@don-ew minor comment.

apps/src/types/types.ts
don-ew: `postId` I think is better since it's in line with `post_id` from WP.

mahieddineak: @don-ew good, I will update the API response and only use camel case.

mahieddineak (approved):
@don-ew all good!

PR #351, opened 2025-03-26
refs #81965: Climate variable config and class (map update)

don-ew:
## Description

- This is a continuation of https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/350
- Threshold dropdown now only displays when multiple threshold values are available
- Refactored map-related components for more modular approach to handle different map types
  - Map components now needs to be returned by the Climate Variable class
- Optimized map layer URL generation


mahieddineak (approved):
@don-ew looks good to me, thanks!

PR #354, opened 2025-03-28
refs #81968: Fix typography issues in the app

alinademi:
## Description

In both the map and the download apps, the typography wasn't applied properly.
There existed console errors such as:

```js
GET https://dev-en.climatedata.ca/CDCSans-Light--aGvsYLY.woff2 net::ERR_ABORTED 404 (Not Found)
```

Examined different ways to load the fonts via `vite.config.ts` including:
- Adding base to the vite config to match the url where app is served
- Leveraging vite\'s publicDir to load the fonts
- Using relative path to load the fonts instead of alias
- directly importing the fonts in `App.ts`

None of the above worked.

The next step was to load a separate font file via vite as a separate entry in addition to explicitly adding it to the templates via the `cdc_app_asset_load()` function. That did not work either.

The last resort was to by-pass all the build processes and load the fonts directly in the templates from the theme\'s assets and to be written directly in the templates head element which resolved the issue.

## How to test:

Navigate to `https://dev-en.climatedata.ca/map-app/` and open the browser console, the errors regarding the missing fonts are gone and font styles are being applied to the app as per the design comps.

In the same way, the standalone app also works since the font assets there are being managed directly by vite.


mahieddineak (changes requested):
@alinademi I left you a suggestion, thanks for looking into it.

alinademi:
> @alinademi thanks for the detailed explanation. I had a quick review and I note that in the file `/apps/src/Global.css` there are already `@font-face` definitions, which might lead to duplicated CSS code. Once changing the URL paths as following, the fonts will load properly as of my testing...

@mahieddineak I did test this before, however, I did not get the same results as you got in my tests. The build command results in:

```console
./../../resources/fonts/CDCSerif-Medium.woff2 referenced in ./../../resources/fonts/CDCSerif-Medium.woff2 didn't resolve at build time, it will remain unchanged to be resolved at runtime
```

As a result, the fonts will be available in `https://dev-en.climatedata.ca/map-app/` but will break the standalone app in localhost which we are using for dev. As for the code in global css I left it for the dev non WP version to work. 

Dev:
image

Please verify this in your test.

mahieddineak:
@alinademi that is fine, as our intention is for the fonts to load in the app on WP page, this app isn't meant to be standalone.

alinademi:
Closing and transferring work to #359

PR #356, opened 2025-03-28
#81979: Add variable post endpoint

mahieddineak:
## Description

This PR adds the necessary WP REST API endpoint to retrieve variable single post data used on the variable info panel, such as:
* Variable title
* Variable associated dataset term name.
* Variable tagline (variable card description).
* Full description
* Technical description
* Relevant sectors
* Relevant trainings list of links
* Featured image

This PR also includes some code refactoring and logic improvements.

**Note**: This PR needs to be merged after the merge of PR #357.

## Related ticket

https://rm.ewdev.ca/issues/81979


fw-child/resources/functions/rest-v3/variable.php
```diff
@@ -0,0 +1,238 @@
...
+
+	if ( ! empty( $training_post_ids ) ) {
+		foreach ( $training_post_ids as $training_post_id ) {
+			// Training post links.
+			$training_link_en = get_permalink( $training_post_id );
+			if ( strpos( $training_link_en, '-en' ) !== false ) {
+				$training_link_fr = str_replace( '-en', '-fr', $training_link_en );
+			} else {
+				$training_link_fr = str_replace( 'climatedata', 'donneesclimatiques', $training_link_en );
+			}
```
xfrenette-crim, line 201: I think this strategy to translate the URL of a training resource is wrong. It seems that it either tries to replace `-en` (as in `dev-en`?) with `-fr`, else it tries to replace `climatedata` with `donneesclimatiques` (as in the domain name?).

But it's not going to work.

First, simply looking for `-en` and replacing it with `-fr` can have unexpected behaviour. Example, the (not real) URL `/resource/new-enhanced-scenario` would be "translated" to `/resource/new-frhanced-scenario` which is an invalid URL.

Second, the URL of ressources are generally not the same in French and in English (it's not just the domain name that changes).

I may be wrong (so make sure to test), but you probably want to use the `translate_permalink()` function (in `framework/resources/functions/setup/rewrite.php`) to get the translated URL of a resource.

xfrenette-crim (changes requested):
Reason of change requests:
- The permalink translation algorithm seems wrong.

Also make sure to change the destination branch

mahieddineak (commented):
@xfrenette-crim all your feedback has been addressing, thanks for looking into that.

xfrenette-crim (approved):

PR #360, opened 2025-04-01
refs: #81974: output breadcrumb chain dynamically

alinademi:
## Description

Output breadcrumb chain dynamically based on the api response.

~~I am adding a [fallback if terms in response is empty](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/360/files#r2022160232). Needs double check. does not seem ideal. Since I am still in the grey about the whole data shape and flow of the app 🤷🏼, I am not sure if it is better to just render empty or the title as per the api. Since if the API response would be:~~

~~variable-datasets: {~~
  ~~terms: [{~~
    ~~term_id: 1,~~
    ~~title: {~~
      ~~en: "Historical Canadian Climate Normals",~~
      ~~fr: "Normales..."~~
    ~~}~~
  ~~}]~~
~~}~~
~~

~~then the code will pick that up and show. Of course it will be more clear to me after implementing the ticket #81992 where I will be dealing with all these variables directly.~~
~~Appreciate any insight.~~

@don-ew @mahieddineak , thanks for the insights, implemented all the comments.

**Here is how the app behaves now:**

**Initial load**: No title is set in the breadcrumbs and it shows empty.

Screenshot 2025-04-05 at 6 43 22 PM

**Dataset selected**: Data set name sets in state and shows in the UI:

image

**Variable selected**: Same as data set. And the variable name shows in the UI.

image

**Locale support**: The breadcrumb supports locale. I have made some improvements for better detection of client locale with `navigator.language`

Screenshot 2025-04-05 at 5 23 34 PM

Then:

image

### DIFF
--- a/apps/src/components/map-header.tsx
+++ b/apps/src/components/map-header.tsx
@@ -125,7 +125,7 @@ const Breadcrumbs: React.FC<{ title: string; onClick: () => void }> = ({
 	const variableTitle = useMemo(() => {
 		if (!climateVariable) return title;
 
-		return climateVariable.getTitle() || __(climateVariable.getId()) || title;
+		return climateVariable.getTitle() || title;
 	}, [climateVariable, __, title]);
 
 	return (
--- a/apps/src/components/map-header.tsx
+++ b/apps/src/components/map-header.tsx
@@ -126,7 +126,7 @@ const Breadcrumbs: React.FC<{ title: string; onClick: () => void }> = ({
 		if (!climateVariable) return title;
 
 		return climateVariable.getTitle() || title;
-	}, [climateVariable, __, title]);
+	}, [climateVariable, title]);
 
 	return (
 		<div className="flex items-center gap-2">
--- a/apps/src/components/map-header.tsx
+++ b/apps/src/components/map-header.tsx
@@ -22,6 +22,7 @@ import { useAnimatedPanel } from '@/hooks/use-animated-panel';
 import { MapInfoProps, ProviderPanelProps } from '@/types/types';
 import { useLocale } from '@/hooks/use-locale';
 import { useAppSelector } from "@/app/hooks";
+import { useClimateVariable } from "@/hooks/use-climate-variable";
 
 /**
  * MapHeader component, displays the header for the map with breadcrumbs and buttons for extra information.
@@ -112,29 +113,34 @@ const Breadcrumbs: React.FC<{ title: string; onClick: () => void }> = ({
   const { __ } = useI18n();
   const { locale } = useLocale();
   const dataset = useAppSelector((state) => state.map.dataset);
-
-  console.log(dataset);
+  const { climateVariable } = useClimateVariable();
 
   const datasetName = useMemo(() => {
     if (!dataset) return '';
-
     return locale === 'fr' && dataset.title.fr
       ? dataset.title.fr
       : dataset.title.en;
   }, [dataset, locale]);
 
+  const variableTitle = useMemo(() => {
+    if (!climateVariable) return title;
+
+    return __(climateVariable.getId()) || title;
+
+  }, [climateVariable, __, title]);
+
   return (
     <div className="flex items-center gap-2">
       {datasetName && <span>{datasetName}</span>}
-      {datasetName && title && <span>/</span>}
-      {title && (
+      {datasetName && variableTitle && <span>/</span>}
+      {variableTitle && (
         <Button
           variant="ghost"
           className="text-md text-cdc-black hover:text-dark-purple hover:bg-transparent p-0 h-auto"
           onClick={onClick}
           aria-label={__('View details')}
         >
-          {title}
+          {variableTitle}
           <Info />
         </Button>
       )}
--- a/apps/src/components/map-header.tsx
+++ b/apps/src/components/map-header.tsx
@@ -107,45 +107,44 @@ MapHeader.displayName = 'MapHeader';
  * Breadcrumbs component, displays the breadcrumbs for the selected state of the map.
  */
 const Breadcrumbs: React.FC<{ title: string; onClick: () => void }> = ({
-  title,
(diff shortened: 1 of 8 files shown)


apps/src/components/map-header.tsx
```diff
@@ -104,27 +105,61 @@ MapHeader.displayName = 'MapHeader';
...
+        const datasetTerms = currentVariable?.meta?.taxonomy?.['variable-datasets']?.terms;
+
+        if (datasetTerms?.[0]) {
+          const firstDataset = datasetTerms[0];
+          setDatasetName(locale === 'fr' && firstDataset.title.fr
+            ? firstDataset.title.fr
+            : firstDataset.title.en);
+        } else {
+          // Fallback to default dataset name if no terms found
+          setDatasetName(__('Historical Canadian Climate Data'));
```
alinademi, line 136: // Fallback to default dataset name??

don-ew: Personally, I don't think there should be a fallback. I'd rather display nothing if there's no variable loaded.

apps/src/components/map-header.tsx
```diff
@@ -104,27 +105,61 @@ MapHeader.displayName = 'MapHeader';
...
-				<Info />
-			</Button>
-		</div>
-	);
+  const { __ } = useI18n();
+  const { locale } = useLocale();
+  const [datasetName, setDatasetName] = useState<string>('');
+
+  useEffect(() => {
+    const fetchDatasetName = async () => {
```
don-ew, line 116: @alinademi the dataset name should be set in the state after selecting the dataset from the selection. Instead of requesting from the API again, we should use the values from the state.

apps/src/types/types.ts
```diff
@@ -572,3 +572,70 @@ export interface ChoroValuesOptions {
 	emissionScenario: string;
 	decimals: number;
 }
+
```
don-ew, line 575: @alinademi may not be necessary since we already have a type called `ApiPostData`. Please reuse that or enhance it as needed.

https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/blob/8546d0ad3ba54dbc9d2f2897e9586552d1121b6f/apps/src/types/types.ts#L93

apps/src/components/map-header.tsx
```diff
@@ -104,27 +105,61 @@ MapHeader.displayName = 'MapHeader';
  * Breadcrumbs component, displays the breadcrumbs for the selected state of the map.
  */
 const Breadcrumbs: React.FC<{ title: string; onClick: () => void }> = ({
-	title,
-	onClick,
+  title,
+  onClick,
```
don-ew, line 109: @alinademi could you also please use the same indent settings as the code? As you may have noticed, the code is using tabs that's 2 spaces size, for consistency.

don-ew (changes requested):

apps/src/components/map-header.tsx
mahieddineak: Definitely, there is no fallback; if a variable doesn't have a dataset, having a default, non-relevant dataset will be considered inaccurate information. The empty/non dataset is the correct behavior.

apps/src/components/map-header.tsx
```diff
@@ -104,27 +105,61 @@ MapHeader.displayName = 'MapHeader';
...
+          setDatasetName(locale === 'fr' && firstDataset.title.fr
+            ? firstDataset.title.fr
+            : firstDataset.title.en);
+        } else {
+          // Fallback to default dataset name if no terms found
+          setDatasetName(__('Historical Canadian Climate Data'));
+        }
+      } catch (error) {
+        console.error('Error fetching dataset name:', error);
+        setDatasetName(__('Historical Canadian Climate Data'));
```
mahieddineak, line 140: @alinademi same comment as above.

apps/src/components/map-header.tsx
```diff
@@ -104,27 +105,61 @@ MapHeader.displayName = 'MapHeader';
...
+      }
+    };
+
+    fetchDatasetName();
+  }, [locale, title, __]);
+
+  return (
+    <div>
+      <span className="me-2">{__('Canada')}</span>/
```
mahieddineak, line 149: Please make sure the breadcrumbs follow the structure on Figma as follows:
Dataset name > Variable name

apps/src/types/types.ts
```diff
@@ -572,3 +572,70 @@ export interface ChoroValuesOptions {
...
+  meta?: {
+    content: {
+      title: MultilingualField;
+      thumbnail?: string;
+    };
+    taxonomy?: {
+      sector: { terms: TermItem[] };
+      region: { terms: TermItem[] };
+      'var-type': { terms: TermItem[] };
+      'variable-datasets': { terms: TermItem[] };
```
mahieddineak, line 622: No s, 'variable-dataset'

apps/src/types/types.ts
```diff
@@ -572,3 +572,70 @@ export interface ChoroValuesOptions {
...
+  'updated-on': string;
+  content: {
+    title: MultilingualField;
+    thumbnail?: string;
+  };
+  taxonomy: {
+    sector: { terms: TermItem[] };
+    region: { terms: TermItem[] };
+    'var-type': { terms: TermItem[] };
+    'variable-datasets': { terms: TermItem[] };
```
mahieddineak, line 589: No s, 'variable-dataset'

mahieddineak (commented):
@alinademi I left you a couple of comments.

apps/src/components/map-header.tsx
alinademi: Makes total sense.

apps/src/components/map-header.tsx
alinademi: noted

don-ew (approved):

PR #364, opened 2025-04-03
refs #000: Added .editorconfig for formatting consistency.

don-ew:
## Description

- Adds an editorconfig for formatting consistency such as indent style


apps/.editorconfig
mahieddineak, line 1: @don-ew I suggest combining items as follow:
```
# JavaScript/TypeScript settings
[*.{js,ts,jsx,tsx}]
quote_type = single
```

PR #365, opened 2025-04-03
refs #000: Fixed issue caused by change in the API response.

don-ew:
## Description

Updates how the terms are extracted from the response


mahieddineak (approved):
LGTM

PR #366, opened 2025-04-03
refs #000: Fixed issue with the variable response

don-ew:
## Description

This fixes the issue with the variable list crashing, which was due to the API updates

### DIFF
--- a/apps/src/services/services.ts
+++ b/apps/src/services/services.ts
@@ -141,10 +141,12 @@ export const fetchTaxonomyData = async (
  * Fetches posts data from the API
  * @param postType - the post type to fetch
  * @param filters - the filters to apply to the data
+ * @param section - the section we are in
  */
 export const fetchPostsData = async (
 	postType: string,
-	filters: Record<string, string | number | null>
+	filters: Record<string, string | number | null>,
+	section: string
 ): Promise<ApiPostData[]> => {
 	// check if the there is a dummy response for the slug, which means the API is not yet implemented for it in the backend
 	const fetchFromApi = !Object.keys(dummyResponses).includes(postType);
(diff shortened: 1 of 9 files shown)


apps/src/services/services.ts
```diff
@@ -139,18 +141,26 @@ export const fetchTaxonomyData = async (
  * Fetches posts data from the API
  * @param postType - the post type to fetch
  * @param filters - the filters to apply to the data
+ * @param section - the section we are in
```
mahieddineak, line 144: @don-ew could we use the term "app" instead of "section"? And add in the comments the possible options: map and download?

apps/src/services/services.ts
```diff
@@ -139,18 +141,26 @@ export const fetchTaxonomyData = async (
...
+	section: string
 ): Promise<ApiPostData[]> => {
 	// check if the there is a dummy response for the slug, which means the API is not yet implemented for it in the backend
 	const fetchFromApi = !Object.keys(dummyResponses).includes(postType);
 
 	// fetch from the API or directly return dummy json response
 	const data: ApiPostData[] = fetchFromApi
 		? await fetch(
-				`${WP_API_DOMAIN}/wp-json/cdc/v3/${postType}-list`
+				`${WP_API_DOMAIN}/wp-json/cdc/v3/${postType}-list?app=${section}&variable_dataset_term_id=${postType}`, {
```
mahieddineak, line 157: The variable_dataset_term_id is set incorrectly; it should be filter.

apps/src/services/services.ts
```diff
@@ -139,18 +141,26 @@ export const fetchTaxonomyData = async (
  * Fetches posts data from the API
  * @param postType - the post type to fetch
  * @param filters - the filters to apply to the data
+ * @param section - the section we are in
  */
 export const fetchPostsData = async (
 	postType: string,
-	filters: Record<string, string | number | null>
+	filters: Record<string, string | number | null>,
```
mahieddineak, line 148: The filters should represent the dataset term ID, and it is only integer other than 0.

mahieddineak (commented):
@don-ew I left you some comments about the logic. Another minor comment: if it is a quick fix, it is possible to gray out the variable button if the dataset isn't selected? Instead of opening it and seeing nothing.

apps/src/services/services.ts
don-ew: It's used section for the most part except for the API call. I think we should stick to this.

apps/src/services/services.ts
don-ew: Fixed

apps/src/services/services.ts
don-ew: We can revisit this later but this is not used for the term ID. It's used for the other filters (e.g. section, type).

PierrePaul (approved):

PR #375, opened 2025-04-09
Refs: #81974 Fix download app error with missing `LocaleProvider`

alinademi:
## Description

Fixed the error "`useLocale` must be used within a `LocaleProvider`" that was occurring in the download app after recent changes to the locale context. The map app was working correctly because it had `LocaleProvider` in its `App.tsx` component, but the download app was missing this provider entirely from the start.

image

Changes:
- Added `LocaleProvider` to download app's component hierarchy in `app.tsx`
- Removed redundant `LocaleProvider` from `main-download.tsx`
- Maintained the `I18nProvider` in the main entry point

The download app now correctly follows the same provider architecture as the map app, with `LocaleProvider` contained within the `App` component rather than in the main entry file.

## Results

Both download-app and map-app are working now.

image

image


mahieddineak (approved):
@alinademi perfect, thanks!

PR #377, opened 2025-04-10
refs #81986: add marker pin and recent location when clicking interactive map

andruxnet:
## Description

This PR implements adding a marker to the map when clicking on an area of the interactive map, and also to add as a recent location.


mahieddineak (approved):
Thanks @andruxnet, looks good to me!

PR #378, opened 2025-04-10
Refs: #81993: Add variable selection search filter

alinademi:
## Description

Implements variable search functionality

Added search and filter capabilities to the variables list by implementing a
debounced search hook and creating a search filter component. This allows users
to quickly find climate variables by typing in search terms without having to
scroll through the entire list.

The implementation includes:
- Created VariableSearchFilter component with search input and clear button
- Integrated search functionality in the variables sidebar component
- Updated climate-variable-slice with new search query state management
- Added highlight for matched search terms in radio card components
- Added useDebounce custom hook to prevent excessive state updates

https://github.com/user-attachments/assets/b0b34e81-fd52-4584-9bb3-d1213ef43b28

### DIFF
--- a/apps/src/components/variable-search-filter.tsx
+++ b/apps/src/components/variable-search-filter.tsx
@@ -9,12 +9,14 @@ import { Search, X } from 'lucide-react';
 import { useAppDispatch, useAppSelector } from '@/app/hooks';
 import { setSearchQuery, selectSearchQuery } from '@/store/climate-variable-slice';
 import { useDebounce } from '@/hooks/use-debounce';
+import { useI18n } from '@wordpress/react-i18n';
 
 export const VariableSearchFilter: React.FC = () => {
   const dispatch = useAppDispatch();
   const globalSearchQuery = useAppSelector(selectSearchQuery);
   const [localSearchQuery, setLocalSearchQuery] = useState(globalSearchQuery);
   const debouncedSearchQuery = useDebounce(localSearchQuery, 300);
+  const { __ } = useI18n();
 
   // Update Redux state when debounced value changes
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


apps/src/components/variable-search-filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+    dispatch(setSearchQuery(''));
+  };
+  return (
+    <div className="relative w-full py-3">
+      <div className="absolute inset-y-0 left-0 flex items-center pl-3 pointer-events-none">
+        <Search className="h-5 w-5 text-gray-400" />
+      </div>
+      <input
+        type="text"
+        placeholder="Search"
```
mahieddineak, line 41: This should be translatable.

mahieddineak (commented):
@alinademi one minor comment.

apps/src/components/variable-search-filter.tsx
alinademi: Done

alinademi:
OK, so I managed to make all the new requirements using the existing pattern and debounce that was already in place, fairly quickly. I also added some accessibility improvements.

https://github.com/user-attachments/assets/fd38410d-159b-4d1c-a8cb-d0524bf34855

We now have:
- Client-side filtering using a memoized function to efficiently filter
the list based on the search query. This ensures recalculations occur only when
necessary, improving performance.
  -  a clear message when no results are found and
  -  the count of filtered results.
  -  a smooth fade-in animation for new cards appearing in the filtered results.
  -  Unfortunately could not use the profiler in react-devtools but the browser performance tab, shows a good amount of throttling and nice performance once the cards are being filtered. There are improvements to be done for animation performance, like using a custom hook that detects the firs render, etc, but my tests show it is already fine for the scope the way it already is.


- Accessibility of the variable search filter:
  - proper ARIA roles and labels
  - Icons are marked as aria-hidden to avoid unnecessary announcements.

- Keyboard navigation by maintaining Escape key handling to clear the
search and ensuring proper focus management after clearing.
@mahieddineak @don-ew

mahieddineak (approved):
@alinademi amazing job, approved!

PR #381, opened 2025-04-10
refs #82001/#82002: Dynamic colour generation

don-ew:
## Description

- Map colour options can now be selected
- Map is updated when the colour is selected
- Fixed an issue where the map is not rendered when the "Annual" option is not available
- Hide Absolute/Delta field when the variable doesn't support it
- Allow users to select between Continuous and Discrete
- Map colour options can be overridden via config (via customColourSchemes)


mahieddineak (approved):
@don-ew Great work, thanks!

PR #383, opened 2025-04-11
81976 map feature charts panel

romainnugou:
## Description

Chart panel

- update panel header info (with dataset name, climate variable name and version)
- include median, relative to baseline and range values (from the same component used in location modal)
- making chart dynamic using the "generate-charts" API provided by dataclimatedata.crim.ca
- add the export options (PDF, PNG, CSV and print)

- Fix the relative to baseline bug we had on the location modal for climate variable with DoY unit.
- Fix the empty frequency issue


apps/src/components/climate-data-chart.tsx
```diff
@@ -353,21 +389,33 @@ const ClimateDataChart: React.FC<{ title: string; data: ClimateDataProps }> = ({
...
+								return `+${formattedValue} ${unit}`;
+							}
+							return `${formattedValue} ${unit}`;
+						}
+					};
+
+					const points = (this as unknown as { points?: TooltipPoint[] }).points;
+					return points?.map((point: TooltipPoint) => {
+						if (point.series.type === 'arearange') {
+							return `<span style="color:${point.series.color}">●</span> ${point.series.name}: <b>${formatValue(point.low)}</b> - <b>${formatValue(point.high)}</b><br/>`;
```
mahieddineak, line 413: Could we change ● to its associated HTML encoding entity such as `&bull;` or `#8226;`?

apps/src/components/climate-data-chart.tsx
```diff
@@ -380,33 +428,57 @@ const ClimateDataChart: React.FC<{ title: string; data: ClimateDataProps }> = ({
...
-					// formatter: function () {
-					// 	const value = this.axis.defaultLabelFormatter.call(this);
-					// 	if (localized_chart_unit === 'doy') {
-					// 		return value_formatter(value, options.chart.unit, var_fields.decimals, false, options.lang);
-					// 	} else {
-					// 		return value + ' ' + localized_chart_unit;
-					// 	}
-					// }
+						const unit = climateVariable?.getUnit();
+						if(unit === "doy") {
```
mahieddineak, line 432: I would suggest using switch instead of multiple if and if else, just in case we have more edge cases.

apps/src/components/map-layers/raster-precalculated-climate-variable-values.tsx
```diff
@@ -132,28 +143,26 @@ const RasterPrecalcultatedClimateVariableValues: React.FC<RasterPrecalcultatedCl
...
 
 	// Value formatter (for delta, for units)
 	const valueFormatter = (value: number, delta: boolean = (dataValue === 'delta')) => {
+		const unit = climateVariable?.getUnit();
 		let str = '';
 
-		// If delta, we add a "+" for positive values
-		if(delta && value > 0) {
-			str += '+';
+		if(unit === 'doy') {
```
mahieddineak, line 153: Same comment as above.

mahieddineak (approved):
@romainnugou approved with minor comments.

romainnugou:
@mahieddineak I fixed what you commented in this PR.

I'm now working on the chart config (as asked in the RM task).

PR #385, opened 2025-04-12
refs: #81996: fix scenario comparison feature, remove unused scenario options from map-slice

andruxnet:
## Description

This PR includes the following missing emission scenario features:

- Use correct scenario data for rendering the map
- Add current scenario label at the bottom of the map
- Fix comparison map
- Make sure the same scenario does not appear in both scenario dropdowns when  comparing

Also, old logic for emission scenarios was removed in favor of new class structure.


mahieddineak (approved):
Thanks @andruxnet

PR #389, opened 2025-04-15
refs #81984: fix location name in modal.

romainnugou:
## Description

Fix location name in modal.
- For grid cells: stay the same, we use get_location_by_coords API
- For others (census, health regions, watersheds): we use the title we have in map layer


mahieddineak (approved):
Thanks @romainnugou

PR #393, opened 2025-04-17
81972-dynamic url filters

alinademi:
## Description

Adds map filters to url as query params


apps/src/features/map/map-slice.ts
```diff
@@ -82,6 +82,9 @@ const mapSlice = createSlice({
 		setVariable(state, action: PayloadAction<string>) {
 			state.variable = action.payload;
 		},
+		setFrequency(state, action: PayloadAction<string>) {
```
don-ew, line 85: Not sure if we need this.

The frequency should be set on the climate variable instead.

alinademi: Done and some!

apps/src/hooks/use-url-sync.ts
```diff
@@ -0,0 +1,282 @@
...
+	const timePeriodEnd = useAppSelector((state) => state.map.timePeriodEnd);
+	const variable = useAppSelector((state) => state.map.variable);
+	const decade = useAppSelector((state) => state.map.decade);
+	const thresholdValue = useAppSelector((state) => state.map.thresholdValue);
+	const interactiveRegion = useAppSelector(
+		(state) => state.map.interactiveRegion
+	);
+	const frequency = useAppSelector((state) => state.map.frequency);
+	const mapColor = useAppSelector((state) => state.map.mapColor);
+	const opacity = useAppSelector((state) => state.map.opacity);
```
don-ew, line 37: Most of this should be taken from the climate variable state instead of the map state.

don-ew (changes requested):

apps/src/hooks/use-url-sync.ts
alinademi: I prioritized climate variable state in the last commit.
I am a little confused about the dual state that we are managing here. 
Here are the inconsistencies between map state and climate variable state for the same filter names:

- Variable Name
Climate variable: `var` (example: `var=tx_max`)
Map state: `mapVar` (example: `mapVar=tx_max`)

- Interactive Region
Climate variable: `region` (example: `region=gridded_data`)
Map state: `mapRegion` (example: `mapRegion=gridded_data`)

- Time Period/Date Range
Climate variable: `dateRange` (example: `dateRange=2040,2070`)
Map state: `period` (example: `period=2070`) - only stores the end year

- Frequency
Climate variable: `freq` (example: `freq=ann`)
Map state: `mapFreq` - though this parameter doesn't appear in your current URL implementations

- Color Scheme
Climate variable: `clr` (example: `clr=default`)
Map state: `color` (example: `color=default`)

- Threshold
Climate variable: `th` (example: `th=tx_max`) - often a string value
Map state: `threshold` (example: `threshold=5`) - typically a numeric value

Now, some components read directly from map state
Some features only exist in map state and not in climate variable state
For parameters that exist in both states, we need to update both to prevent inconsistencies

For example, when we load a URL with `dateRange=2040,2070`, we need to:

- Update the climate variable state with that date range
- Calculate and update the map state's `timePeriodEnd` value to [2070]

Or, if a URL has `region=watershed`, we need to:

- Update the climate variable state with that region
- Update the map state's `interactiveRegion` value

Why are we maintaining two states? why don't we have something like a middleware thunk or a state sync util that takes care of this?

don-ew (approved):

PR #396, opened 2025-04-21
refs #82007: reset fields when navigating back to previous steps

andruxnet:
## Description

This PR will reset fields when navigating back to previous steps in the download app.


apps/src/components/download/step-send-request.tsx
```diff
@@ -119,8 +128,8 @@ const StepSendRequest = React.forwardRef((_, ref) => {
 						className="sm:w-64 mb-2"
 						placeholder={__('john.doe@gmail.com')}
 						value={email}
-						onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
-							dispatch(setEmail(e.target.value));
+						onChange={(e) => {
+							setRequestFieldValue('email', e.target.value);
```
don-ew, line 132: Hi @andruxnet for `email` and `subscribe` fields, I think they should be more a property of the download app itself as opposed to the climate variable. As such, please store these info in the download state instead. Thanks!

don-ew (changes requested):

PR #398, opened 2025-04-21
refs #85551: Display years and (Change from 1971-2000) on tooltip.

romainnugou:
## Description

Fix chart tooltip bug on years
- add year on "Annual values" tooltip
- add "(Change from 1971-2000)" on "30 year changes" tooltip


mahieddineak (approved):
@romainnugou LGTM, thanks!

PR #399, opened 2025-04-22
refs #85554: Fix chart CSV exports for 30 year averages and 30 year c…

romainnugou:
## Description

Fix chart CSV exports for 30 year averages and 30 year changes options.


mahieddineak (approved):
@romainnugou LGTM, thanks!

PR #403, opened 2025-04-22
refs #85610: Fix scroll on chart to increase or decrease range.

romainnugou:
## Description

Fix scroll on chart to increase or decrease range.


mahieddineak (approved):
Thanks @romainnugou!

PR #405, opened 2025-04-23
Refs: 82011: Add variable selection search filter to download app

alinademi:
## Description

Introduces a new `app` prop to the `VariableSearchFilter` component to differentiate behavior between the "map" and "download" contexts.

Also does some lining as some files are old and are not based on the linting config of the codebase.

## Related ticket

[81993](https://rm.ewdev.ca/issues/81993)
[82010](https://rm.ewdev.ca/issues/82010)

### DIFF
--- a/apps/src/components/download/step-variable.tsx
+++ b/apps/src/components/download/step-variable.tsx
@@ -11,17 +11,18 @@ import {
 import TaxonomyDropdownFilter from '@/components/taxonomy-dropdown-filter';
 import VariableRadioCards from '@/components/variable-radio-cards';
 import VariableFilterCount from '@/components/sidebar-menu-items/variables';
+import { VariableSearchFilter } from '@/components/variable-search-filter';
 
-import { useClimateVariable } from "@/hooks/use-climate-variable";
-import { StepComponentRef } from "@/types/download-form-interface";
-import SectionContext from "@/context/section-provider";
+import { useClimateVariable } from '@/hooks/use-climate-variable';
+import { StepComponentRef } from '@/types/download-form-interface';
+import SectionContext from '@/context/section-provider';
 import useFilteredVariables from '@/hooks/use-filtered-variables';
 
 /**
  * Variable step
  */
 const StepVariable = React.forwardRef<StepComponentRef>((_, ref) => {
-    const { climateVariable, selectClimateVariable  } = useClimateVariable();
+    const { climateVariable, selectClimateVariable } = useClimateVariable();
     const [varType, setVarType] = useState<string>('');
     const [sector, setSector] = useState<string>('');
     const section = useContext(SectionContext);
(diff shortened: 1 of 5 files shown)


apps/src/components/download/step-variable.tsx
```diff
@@ -110,17 +113,23 @@ const StepVariable = React.forwardRef<StepComponentRef>((_, ref) => {
                 />
             </div>
 
-            <div className="w-full mb-4">
-                {renderFilterCount()}
+            <div className="mb-4">
+                <VariableSearchFilter app="download" />
```
don-ew, line 117: @alinademi you can use the section context to determine if we are in the `map` or in the `download`.

```
const section = useContext(SectionContext);
```

`section` will contain either `map` or `download`.

You can use this directly in the `VariableSearchFilter` component and avoid adding extra props.

apps/src/components/variable-search-filter.tsx
```diff
@@ -10,16 +10,24 @@ import { useAppDispatch, useAppSelector } from '@/app/hooks';
...
-import { setVariableListLoading } from '@/features/map/map-slice';
+import { setVariableListLoading as setMapVariableListLoading } from '@/features/map/map-slice';
+import { setVariableListLoading as setDownloadVariableListLoading } from '@/features/download/download-slice';
 
-export const VariableSearchFilter: React.FC = () => {
+interface VariableSearchFilterProps {
+  app?: 'map' | 'download';
+}
+
+export const VariableSearchFilter: React.FC<VariableSearchFilterProps> = ({ app = 'map' }) => {
```
don-ew, line 20: May not be necessary. We can use

```
const section = useContext(SectionContext);
```

don-ew (changes requested):
Hi @alinademi I've added a couple of feedbacks, specifically about the use of the `SectionContext`, which you could use to know whether or not we're in a the map or in the download.

Also, I'm still noticing some extra changes (spacing, newlines) that are more about code formatting. This adds some "noise" to the CR. 

Perhaps your IDE is auto-formatting the things?

apps/src/components/download/step-variable.tsx
alinademi: Done.

apps/src/components/variable-search-filter.tsx
alinademi: Done

apps/src/components/download/step-variable.tsx
alinademi: Re IDE formatting, everything is off as you see in the video. I mentioned in the PR comments above:
>Also does some lining as some files are old and are not based on the linting config of the codebase.

Appreciate it if you could tell me what are your settings and formatter so that we can match. since we have prettier config in the repo, as well as editor config.

**I suggest everyone uses the same formatter (best of all prettier) so that this does not happen.**

OR

**I can add a CI script or a githook that automatically formats with prettier on commit for everyone**

I am open to suggestions.

I turned off every formatter (even default typescript) since you made comment in some older PRs and that is so inconvenient.

https://github.com/user-attachments/assets/0c3f66b9-3e6e-4439-87dd-a842e1227982

apps/src/components/download/step-variable.tsx
```diff
@@ -81,7 +83,7 @@ const StepVariable = React.forwardRef<StepComponentRef>((_, ref) => {
       const hasCurrent = currentId != null && filteredList.some(v => v.id === currentId);
       if (!hasCurrent) {
         // select first available variable by default
-        selectClimateVariable(filteredList[0], dataset);
```
don-ew, line 84: @alinademi I'm not sure if we should remove `dataset` here. It's needed to set the dataset in the climate variable class.

don-ew (changes requested):

apps/src/components/download/step-variable.tsx
don-ew: @alinademi in that case, I'd suggest to just be mindful when committing or creating the PR. If there's any line in the code that was changed automatically that you didn't touched and is in no way essential to the task, you could manually revert those specific lines. Example:

Screenshot 2025-04-25 at 9 37 31 AM

For some reason, it's only your PRs that are having these extra line changes, which is why I initially suspected it's something to do with your IDE. For now, just try to manually revert those kinds of changes.

apps/src/components/download/step-variable.tsx
alinademi: done

PR #407, opened 2025-04-25
81967 add sea level variable

romainnugou:
## Description

Add sea level climate variable.

- add variable in config
- add specifics scenario to appConfig
- create a new class for this variable: SeaLevelClimateVariable
- create a new class for the value inside tooltip: SeaLevelClimateVariableValues
- had to adapt TimePeriodsControl to handle Sea level variable behavior about date range (this variable has a different interval, we have to display only the end year, it starts à 2006 but next should be 2010, then 2020, until 2100, and there's a special case inside this special case, date range has to be forced on 2100 on a specific scenario).

### DIFF
--- a/apps/src/components/sidebar-menu-items/time-periods-control.tsx
+++ b/apps/src/components/sidebar-menu-items/time-periods-control.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import * as Slider from '@radix-ui/react-slider';
 import { useI18n } from '@wordpress/react-i18n';
 import { useAppDispatch, useAppSelector } from '@/app/hooks';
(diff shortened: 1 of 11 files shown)


apps/src/components/sidebar-menu-items/time-periods-control.tsx
don-ew, line 1: Hi @romainnugou , as discussed, with all these special handling for the sea-level, it might be better to have a separate component so as not to complicate this. Thanks!

don-ew (changes requested):

apps/src/components/sidebar-menu-items/time-periods-control.tsx
romainnugou: That's fixed.
I created a TimePeriodsControlForSeaLevel component.

don-ew (approved):

PR #410, opened 2025-04-25
refs #81982: Various updates to the standard (raster) map, including fixes

don-ew:
## Description

Various updates to the raster map including fixes:

- Version, scenario, region not updating after switching to a different variable
- Fixed opacity settings lost when switching between tags
- Prevent unnecessary re-rendering of the map
- Added dynamic grid type depending on the version

## Related ticket

- https://rm.ewdev.ca/issues/81982


mahieddineak (approved):
@don-ew LGTM, thanks!

PR #419, opened 2025-04-29
Refs: 81973: Fix - variable auto-select once selecting the sea-level variable was not working

alinademi:
## Description
Issue:
variable auto-select once selecting the sea-level variable was not working.

It wasn't for sea-level only but for any var that was not the first one in the list.

NOTE ⚠️ : This is branched off #416 so must be merged after that.

### DIFF
--- a/apps/src/components/variable-radio-cards.tsx
+++ b/apps/src/components/variable-radio-cards.tsx
@@ -22,7 +22,7 @@ const ResultsMessage: React.FC<{ message: string }> = ({ message }) => (
 	<div className="col-span-2 p-4 text-center">{message}</div>
 );
 
-const VariableRadioCards: React.FC<{
+interface VariableRadioCardsProps {
 	filterValues: Record<string, string>;
 	selected: PostData | null;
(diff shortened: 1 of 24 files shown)


apps/src/components/variable-radio-cards.tsx
```diff
@@ -70,6 +70,13 @@ const VariableRadioCards: React.FC<VariableRadioCardsProps> = ({
 			// Skip if we're already fetching
 			if (fetchingRef.current) return;
 			if (prevDatasetRef.current === dataset.term_id && variableList.length > 0) {
+				// If we already have variables for this dataset but never selected one,
+				// make sure to select the first variable
+				if (!hasAutoSelectedRef.current && !urlParamsLoaded && variableList.length > 0) {
```
mahieddineak, line 75: The `variableList.length > 0` condition already set in the parent `if`.

mahieddineak (approved):
@alinademi very minor comment, I will merge it anyway and once I have a moment on another PR, you can fix it.

PR #422, opened 2025-04-30
Refs: 81973-social share panel -- No social Icons

alinademi:
## Description

- fix the redundant if statement in `variable-radio-cards.tsx`
- solves the huge issue of the not shareability of the URL. This was occurring due to the competition between selecting the first variable in the dataset functionality and being able to update the sidebar controls and eventually the map from the URL(shareability)

TODO:
- [x] Bug: When URL gets shared, the map updates with correct filter but if you open the dataset panel, you'll see that the first dataset is selected(on init) despite the other dataset is correctly in the URL and the variable from it is selected and related layer is showing on the map.
- [x] Select a variable (any variable other than the 1st one in the list), The map will change, Open the “Variables” panel again, and it auto switches to the 1st one in the list
- [x] if I enter the base url `https://dev-en.climatedata.ca/map-app/` only without the query params, then the map does not load.

NOTE ⚠️:
These two will be solved in the next PR as they were not as critical as the other 3
- [ ] Bug: Map header component disappears in initial state
- [ ] Implement social sharing buttons. @don-ew I recommend [React-share](https://www.npmjs.com/package/react-share) package. Since every platform has its own url params, I prototyped something, but soon found that having this dedicated package is less headache.

## Related ticket
https://rm.ewdev.ca/issues/81973

### DIFF
--- a/apps/src/components/map-info/share-map-modal.tsx
+++ b/apps/src/components/map-info/share-map-modal.tsx
@@ -4,7 +4,7 @@
  * A modal component that allows users to copy the map link or share on social media.
  *
  */
-import React, { useState } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { useI18n } from '@wordpress/react-i18n';
 
 // components
@@ -18,6 +18,7 @@ import {
 	ModalSectionBlockDescription,
 } from '@/components/map-info/modal-section';
 import Modal from '@/components/ui/modal';
+import { useAppSelector } from '@/app/hooks';
 
 const ShareMapModal: React.FC<{
 	isOpen: boolean;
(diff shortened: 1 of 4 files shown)


apps/src/hooks/use-url-sync.ts
```diff
@@ -292,89 +291,199 @@ export const useUrlSync = () => {
...
+						config.id === firstVariable.id);
+					
+					if (matchingConfig) {
+						// Set climate variable with dataset context
+						const variableWithDataset = {
+							...matchingConfig,
+							datasetType: firstDataset.dataset_type
+						};
+
+						console.log({
```
mahieddineak, line 360: This has to be removed, no console log.

apps/src/components/map-info/share-map-modal.tsx
```diff
@@ -26,18 +27,30 @@ const ShareMapModal: React.FC<{
 	const [copyLinkVariant, setCopyLinkVariant] = useState<
 		'destructive' | 'secondary'
 	>('destructive');
+	const [currentUrl, setCurrentUrl] = useState<string>('');
```
mahieddineak, line 30: Instead of an empty value, I would suggest the map app URL without parameters.

mahieddineak (approved):
@alinademi all good, 2 minor comments and we can merge.

apps/src/hooks/use-url-sync.ts
alinademi: done!

apps/src/components/map-info/share-map-modal.tsx
alinademi: It is a good addition @mahieddineak 
a very simple way is hardcoding it this way which will be ugly:
```ts
const [currentUrl, setCurrentUrl] = useState<string>(
	typeof window !== 'undefined' ? window.location.href : 'https://dev-en.climatedata.ca/map-app/'
);
```

a nicer way is to have a config file and add `BASE_URL` to it and read based on the environment. However, I thing since this is only when the modal is open and in the `useEffect` on line 39 we already are checking for the availability of the URL, then maybe it is a little overkill. I would love to have a config always to set these things dynamically across the app. I can do it in the next PR if you want me to.

mahieddineak: Let us then keep it aside for now, thanks @alinademi !

PR #426, opened 2025-05-01
refs #85760: remove browser language.

PierrePaul:
## Description

We changed the locale based on the navigator locale. This is not needed, as it is being set by WordPress, depending on the domain.


mahieddineak (approved):
@PierrePaul all good, thanks!

PR #427, opened 2025-05-01
refs #85904: adding a back link to WordPress.

PierrePaul:
## Description

Fix the back link to WordPress on the Map app.
Add the header with the back link on the Download app.


mahieddineak (approved):
@PierrePaul Great!

PR #428, opened 2025-05-01
refs #85554: Update cart csv export, 30y avg and 30y changes exports …

romainnugou:
## Description

Update cart csv export, 30y avg and 30y changes exports are now the same.


mahieddineak (approved):
Thanks @romainnugou!

PR #430, opened 2025-05-02
refs #82015: add selection limit to shape drawing in download map step

andruxnet:
## Description

This PR will add the missing cell selection limit when drawing a shape in the download map step.

### DIFF
--- a/apps/src/components/download/raster-download-map.tsx
+++ b/apps/src/components/download/raster-download-map.tsx
@@ -19,7 +19,12 @@ import { useI18n } from '@wordpress/react-i18n';
 import { useMap } from '@/hooks/use-map';
 import { useAppDispatch, useAppSelector } from '@/app/hooks';
 import { useClimateVariable } from '@/hooks/use-climate-variable';
-import { DownloadType, InteractiveRegionOption } from '@/types/climate-variable-interface';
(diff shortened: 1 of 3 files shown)


apps/src/components/download/raster-download-map.tsx
```diff
@@ -39,13 +44,34 @@ export default function RasterDownloadMap(): React.ReactElement {
 		gridLayerRef.current?.clearSelection();
 	};
 
+	const maxCellsAllowed = useMemo(() => {
+		const freq = climateVariable?.getFrequency();
+		const format = climateVariable?.getFileFormat();
+
+		if (freq === FrequencyType.ALL_MONTHS) {
```
mahieddineak, line 51: @andruxnet At this step, I think all these conditions of frequency and format don't get applied because they shouldn't be selected yet, so I think we can simply limit it to 1000.

mahieddineak (commented):
@andruxnet a minor comment, thanks!

apps/src/components/download/raster-download-map.tsx
andruxnet: Agreed. Change is done.

mahieddineak (approved):

PR #431, opened 2025-05-02
refs #85902: fix time periods slider layout.

romainnugou:
## Description

Fix time periods slider layout
- align "Time periods" title with other titles
- reduce time slider width
- reduce time slider tooltip left and right padding


mahieddineak (approved):
@romainnugou Good trick!

PR #434, opened 2025-05-02
refs #82013: Skip download step 3 when it's a station variable.

romainnugou:
## Description

Skip download step 3 when it's a station variable.


mahieddineak (approved):
@romainnugou LGTM, Thanks!

PR #435, opened 2025-05-02
Refs #80113: Added function to show the marker for user location.

VincenzoGambino:
## Description

- A function was added to show the marker for the user location using the L.Marker method.
- Better handling of the double space.

## Related ticket
https://rm.ewdev.ca/issues/80113


mahieddineak (approved):
@VincenzoGambino LGTM.

PR #447, opened 2025-05-07
refs #86465: Remove seasons option from some variables

don-ew:
## Description

- Forced display of "Seasonal" and "Monthly" options in the Download section for analyzed variables, instead of displaying individual seasons (e.g. Summer) or months.
- The following variable configs have been updated as well:
  - Heating Degree Days
  - Tropical Nights
  - Cooling Degree Days
  - Freeze-Thaw Cycles
  - Wet-days

### DIFF
--- a/apps/src/components/download/step-additional-details.tsx
+++ b/apps/src/components/download/step-additional-details.tsx
@@ -233,9 +233,10 @@ const StepAdditionalDetails = React.forwardRef<StepComponentRef>((_, ref) => {
 				value={climateVariable?.getFrequency() ?? undefined}
 				onValueChange={setFrequency}
 				className={"sm:w-64 mb-4"}
+				downloadType={climateVariable?.getDownloadType() ?? undefined}
 			/>
 
-			{isDownloadTypeAnalyzed
+			{!isDownloadTypeAnalyzed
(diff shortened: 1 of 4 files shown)


mahieddineak (commented):
@don-ew all good, but I think we need to update the `enum FrequencyType` values.

```JS
export enum FrequencyType {
	ANNUAL = "ann",
	ANNUAL_JUL_JUN = "annual_jul_jun",
	MONTHLY = "months",
	SEASONAL = "seasons",
	ALL_MONTHS = "allMonths",
	DAILY = "daily",
}
```

Cursor_and_Analyze_Climate_Data_—_ClimateData_ca

don-ew:
@mahieddineak we can't change those in the `FrequencType` because the code varies depending on the usage and FrequencyType is used in both Map and Download sections (i.e. the code used is different in map than in download).

A better way would be to create a mapping function similar to we map the values here: https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/blob/a890b7fa05b3162a877a40a3e26cb0c6a6223901/apps/src/lib/utils.ts#L58-L57.

mahieddineak (approved):
@don-ew all good, thanks!

PR #460, opened 2025-05-09
Refs: 82021: Implement captcha verification for download requests

alinademi:
## Description

Based on [new context form Mahieddine](https://evolvingweb.slack.com/archives/C07ME79QCT1/p1747165302400149?thread_ts=1747146750.943249&cid=C07ME79QCT1):

The scope is to add the captcha and handle the user interaction with Redux, i.e., store the value in Redux. Refreshing the captcha is also included in the scope of that ticket. On another ticket, when sending or doing a Finch (analyze) submission, the captcha value should be included in the request.

This PR does just that. I removed the verification part from it and let the finch part be done in other tasks.

## How to test:
⚠️ please note that to see the captcha on dev, you need to hardcode the `CAPTCHA_BASE_URL` in `constants.ts` temporarily produce the captcha image.

you would see something like this video (minus the verify button)
https://github.com/user-attachments/assets/89395983-bea3-4468-ba22-4f1b286c520a


apps/src/lib/constants.ts
```diff
@@ -55,6 +55,7 @@ export const WP_API_DOMAIN = (rootElement?.getAttribute('data-wp-home-url')) ??
 
 export const WP_API_VARIABLE_PATH: string = '/wp-json/cdc/v3/variable';
 export const WP_API_LOCATION_BY_COORDS_PATH: string = '/wp-json/cdc/v2/get_location_by_coords';
+export const CAPTCHA_BASE_URL: string = WP_API_DOMAIN + '/site/assets/themes/climate-data-ca/resources/php/securimage/securimage_show.php';
```
alinademi, line 58: hardcoded version would look like this:
```ts
export const CAPTCHA_BASE_URL = 'https://climatedata.ca/site/assets/themes/climate-data-ca/resources/php/securimage/securimage_show.php';
```

don-ew (approved):

PR #466, opened 2025-05-13
refs #86865: Add gridType property to some climate variables configuration

don-ew:
## Description
- Updated the gridType of some climate variables
- Ensure that gridType from configuration is prioritized


mahieddineak (approved):
@don-ew LGTM!

PR #467, opened 2025-05-14
refs #86595: AHCCD precalculated variable

don-ew:
## Description

- Fixes an issue with the summary for station-like variables such as AHCCD precalculated


mahieddineak (approved):
@don-ew all good, a couple of notes that I am not sure if they are in the scope of this project:
* Steps 4 and 5 both have the heading 'Final Step'; I would suggest the step 5 be named Data or Download.
* On the step of selecting stations, could we change "You selected: X Cells" to "You selected: X Stations"?

PR #479, opened 2025-05-15
refs #87268: Fixed a typo in XML layer descriptor

don-ew:
## Description

- Fixed an issue where the map colours are not being applied for the grid interactive region type.


mahieddineak (approved):
LGTM1

PR #480, opened 2025-05-15
refs #82020: add logic for downloading station map files

andruxnet:
## Description

This PR will add logic for downloading files for station maps in the download app.

- Refactor the logic for building stationDownloadUrlsProps in the download flow to use data from climateVariable instead of hardcoded values.
- Add basic for all station based variable types (AHCCD, station-data, climate normals, etc.), some extra logic might need to be added in future tickets.
- Update the fetchStationsList service method to return a consistent station object structure depending on the variable it's used for.
- Limit station type filtering logic only for AHCCD variables, matching the legacy behavior (please confirm this is right).

Note: for the `station_data` variable, the stations list is 8621 items. This makes selecting/unselecting stations slower than for other variables. This same issue happens in v1, but it should be addressed -- likely in a future bug/feature ticket.

### DIFF
--- a/apps/src/components/download/raster-download-map.tsx
+++ b/apps/src/components/download/raster-download-map.tsx
@@ -38,7 +38,7 @@ export default function RasterDownloadMap(): React.ReactElement {
 	);
 	const dispatch = useAppDispatch();
 
-	const [stationTypes, setStationTypes] = useState<string[]>(['T', 'P', 'B']);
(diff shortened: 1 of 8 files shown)


don-ew (changes requested):

mahieddineak (approved):
@don-ew as discussed, I am proceeding with the merge. cc @andruxnet

don-ew (approved):

PR #481, opened 2025-05-15
Refs #81981: Generate map image from server api

VincenzoGambino:
## Description
Added class to the #wrapper-map div to get the screenshot.

## Related ticket
https://rm.ewdev.ca/issues/81981


mahieddineak (approved):
@VincenzoGambino Thanks!

PR #482, opened 2025-05-16
refs #82016: Added default date range property

don-ew:
## Description

- Adds a new property called defaultDateRange, which allows the dateRange to be reset to a default value when moving between steps
- Only display date range in the download summary when the variable is an analyzed variable


mahieddineak (approved):
@don-ew LGTM! I also test it and it is working fine, thanks!

PR #487, opened 2025-05-21
refs #82016: Updated step 5 for Station Data variable

don-ew:
## Description

- Updates the step 5 to support Station Data
- Added a new component to allow users to pick dates


andruxnet (approved):
All good

PR #494, opened 2025-05-22
refs #87696: Update ID mapping to handle 'climate-normals' threshold

don-ew:
## Description

- Fixes an issue in the climate normals variable map where clicking a station shows an empty chart.


andruxnet (approved):
All good 👍

PR #503, opened 2025-05-24
Ref: #87340: Fix: Download: Not all locations (dots) are shown on the map

alinademi:
## Description

Fix: Download: Not all locations (dots) are shown on the map

- Removed redundant window.DATA_URL from
download app
- also fixed bdv variable fetching in services.ts

The merge of https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/507
solved the CORS issue

## Related ticket

https://rm.ewdev.ca/issues/87340

### DIFF
--- a/apps/src/components/download/step-additional-details.tsx
+++ b/apps/src/components/download/step-additional-details.tsx
@@ -60,7 +60,6 @@ const StepAdditionalDetails = React.forwardRef<StepComponentRef>((_, ref) => {
 			if (climateVariable.getDownloadType() === DownloadType.ANALYZED) {
 				const [startYear, endYear] = climateVariable.getDateRange() ?? [];
 				const scenarios = climateVariable.getAnalyzeScenarios() ?? [];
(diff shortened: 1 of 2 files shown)


fw-child/apps/app-download.php
```diff
@@ -73,5 +73,19 @@
     }
 }
 ?>
+
+<script>
```
alinademi, line 77: @don-ew 
To bypass CORS, I had to follow the dirty hack solution that was also implemented before in the [map-app](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/blob/9646a522a351658739efda6fa991aa879d5af842/fw-child/apps/app-map.php#L80)

There are of course more elegant ways, like adding a proper rest endpoint in https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/tree/main/fw-child/resources/functions/rest-v3

But to faster resolve the bug I chose the quick fix.

alinademi: Done!

PR #504, opened 2025-05-24
Refs: #87353: Fix time interval for analyze step

alinademi:
## Description

Fix time interval for analyze step

## Related ticket

https://rm.ewdev.ca/issues/87353

### DIFF
--- a/apps/src/lib/raster-analyze-climate-variable.tsx
+++ b/apps/src/lib/raster-analyze-climate-variable.tsx
@@ -17,7 +17,17 @@ class RasterAnalyzeClimateVariable extends RasterPrecalculatedClimateVariable {
 	}
 
 	getDateRangeConfig(): DateRangeConfig | null {
-		return this.getDatasetType() === "ahccd" ? null : super.getDateRangeConfig();
+		if (this.getDatasetType() === "ahccd") {
+			return null;
+		}
+		
+		return ClimateVariableBase.prototype.getDateRangeConfig.call(this)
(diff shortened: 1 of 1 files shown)


apps/src/lib/raster-analyze-climate-variable.tsx
```diff
@@ -17,7 +17,17 @@ class RasterAnalyzeClimateVariable extends RasterPrecalculatedClimateVariable {
...
+			return null;
+		}
+		
+		return ClimateVariableBase.prototype.getDateRangeConfig.call(this)
+			? ClimateVariableBase.prototype.getDateRangeConfig.call(this)
+			: {
+				min: "1950",
+				max: "2100",
+				interval: 5
+			};
```
don-ew, line 30: @alinademi this will also impact the date slider in the map. Based on the ticket, the map and download should have different date ranges.

I'd suggest having a new method called `getDownloadDateRange`, which gets from the `downloadDateRangeConfig`. 

To minimize the impact, we could also make sure that the **fallback** values coming from `getDownloadDateRange` are retrieved from `getDateRange` in cases where `downloadDateRangeConfig` is not specified.

don-ew (changes requested):

apps/src/lib/raster-analyze-climate-variable.tsx
alinademi: @don-ew Great suggestion. Done!

PR #509, opened 2025-05-26
#81990: Use proper API server URL dynamically

mahieddineak:
## Description

Instead of hardcoding API server URL, use the predefined one on WP.

### DIFF
--- a/fw-child/apps/app-download.php
+++ b/fw-child/apps/app-download.php
@@ -27,6 +27,9 @@ if (
 	$current_lang = $GLOBALS['fw']['current_lang_code'];
 }
 
+// Get the French domain
+$fr_domain = cdc_get_fr_domain();
+
 /**
  * Load translation data for the download application.
  *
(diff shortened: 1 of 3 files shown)


fw-child/apps/app-download.php
```diff
@@ -104,6 +104,12 @@
 	?>
 
     <script>
+        // DATA URL for the download app
+        window.DATA_URL = '<?php echo isset( $GLOBALS['vars']['data_url'] ) ? htmlspecialchars( $GLOBALS['vars']['data_url'], ENT_QUOTES, 'UTF-8' ) : ''; ?>';
+
+        // PAVICS URL for the download app (FINCH)
+        window.PAVICS_URL = '<?php echo isset( $GLOBALS['vars']['pavics_url'] ) ? htmlspecialchars( $GLOBALS['vars']['pavics_url'], ENT_QUOTES, 'UTF-8' ) : ''; ?>';
+
```
xfrenette-crim, line 112: The PAVICS URL might be used _only_ by the PHP, so there would not be a need to add it here. I would remove it. But if eventually the JS needs access to it, we can put it back that way.

xfrenette-crim, line 112: The PAVICS URL might be used _only_ by the PHP, so there would not be a need to add it here. I would remove it. But if eventually the JS needs access to it, we can put it back that way.

xfrenette-crim (approved):

PR #515, opened 2025-05-27
refs #86835: Specify CMIP6 scenarios for the "Days with humidex above threshold"

don-ew:
## Description

- Explicitly define the scenario options for the "Days with humidex above threshold" variable.


andruxnet (approved):
Looks good 👍

PR #549, opened 2025-06-04
refs #86891: make markers sync when comparing scenarios

andruxnet:
## Description

This PR will fix markers not being synced when comparing scenarios.


andruxnet:
No longer needed, replaced by https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/558.

If at all, it may be needed for reference in case I missed something in the new PR.

PR #550, opened 2025-06-04
refs #88419: Updated map threshold option labels.

don-ew:
## Description

- Updates various threshold option labels to include greater/less than signs.


mahieddineak (approved):
LGTM!

PR #555, opened 2025-06-06
refs #88645: Use filenames returned by station data for "Future Building Design"

don-ew:
## Description

- Use the filenames returned by the station data for "Future Building Design" instead of relying on the station name.


andruxnet (approved):
Looks good

PR #561, opened 2025-06-09
Changing the logic on how to download the map

PierrePaul:
## Description

Instead of downloading the blob and inserting it into the DOM, this opens a new tab at the data server that should handle the filename itself.

## Related ticket

https://rm.ewdev.ca/issues/88732


mahieddineak (approved):
LGTM!

PR #565, opened 2025-06-10
refs #88802: fix the ajax call for sea level with ssp 1-2.6.

PierrePaul:
## Description

Send the percentile if it's not already in the scenario name.


mahieddineak (approved):
LGTM

PR #568, opened 2025-06-11
refs #85428: reset compare to value when changing the version.

PierrePaul:
## Description

We were resetting the checkbox value, but not the selected scenario, which ended being filtered if we came back to this scenario.

## Related ticket

https://rm.ewdev.ca/issues/85428


mahieddineak (approved):
LGTM

PR #572, opened 2025-06-12
#88733 adding a delay before the adding the to-raster class

PierrePaul:
## Description

Backend is waiting for the frontend to add the to-raster class before taking a screenshot. The class might be added before the full render is complete, so adding a small delay on the frontend to make sure it is properly rendered.


mahieddineak (approved):
LGTM!

PR #575, opened 2025-06-13
refs #88731: Add temporal threshold config to heating degree days var…

romainnugou:
## Description

Add temporal threshold config to heating degree days variable to enable custom palettes.


mahieddineak (approved):
Thanks @romainnugou! LGTM.

PR #586, opened 2025-06-19
refs #88740: Sea level comparison special case - fix sync issue.

romainnugou:
## Description

Sea level comparison special case - fix sync issue.


mahieddineak (approved):
LGTM! Thanks @romainnugou!

PR #587, opened 2025-06-19
refs #85550: Add variable and threshold in chart name and y axis in c…

romainnugou:
## Description

Add variable and threshold in chart name and y axis in chart exports.


mahieddineak (approved):
@romainnugou LGTM! Thanks.

PR #590, opened 2025-06-20
refs #85550: Fix highcharts error 14 on exporting or changing tab ins…

romainnugou:
## Description

Fix highcharts error 14 on exporting or changing tab inside chart panel.


mahieddineak (approved):
@romainnugou LGTM, thanks!

PR #601, opened 2025-07-02
refs #89369: add grid cells in draw region mode for reference in download map

andruxnet:
## Description

This PR will add a visual of the grid cells for reference in the download map when in draw region mode.


mahieddineak (approved):
Thanks @andruxnet!

PR #602, opened 2025-07-03
refs #88300: Fixed wrong y-axis label showing for "First fall frost" and "Last spring frost"

don-ew:
## Description

- Fixes y-axis label of the chart to show the month and day instead of the number of days while retaining the correct label in the legend.

screenshot_2_1751509337

screenshot_3_1751509362

screenshot_1_1751509300


mahieddineak (approved):
All good, thanks @don-ew

PR #603, opened 2025-07-03
refs #90088: Fixed decimal place not changing

don-ew:
## Description

- Fixes an issue where the changing the decimal place when downloading variable data doesn't do anything.


mahieddineak (approved):
LGTM!

PR #604, opened 2025-07-03
refs #88300: DoY label for delta mode

don-ew:
## Description

- This fixes an issue where the day is formatted as DoY when viewing the 30-year changes instead of just number of days.
- This also fixes an issue where the location modal stays visible after switching variables. This causes some issues related to the data inside of the modal.

Screenshot 2025-07-03 at 2 51 49 PM

Screenshot showing days instead of DoY when in "30-year changes" mode.


apps/src/components/climate-data-chart.tsx
mahieddineak, line 1: @don-ew please make sure all the unit variables are translatable here.

mahieddineak (approved):
LGTM! Thanks @don-ew

PR #669, opened 2026-03-25
CLIM-1287: S2D Location Popup Probability Tooltip

renoirb:
## Original requirement

In the location popup, add an info-popup icon to the probability bars section.

Place the info icon (ℹ) to the right of the probability bars section title (e.g., "Seasonal Mean Temperature Probability").

When activated, the info popup should display a plain-text description of the bar chart data. 

Example:
> The total precipitation has a
> - 33% probability of being unusually high, above 2.1 mm/day
> - 5% probability of being unusually low, below 1.0 mm/day
> 
> relative to the 1991 to 2020 historical climatology.
> 
> The probabilities may not add exactly to 100% due to rounding.

### Post-clarification (2026-04-04)

Client refined the tooltip template — kept the ℹ info icon (no red button), updated the bullet format to use compact cutoff symbols in parentheses:

> The [variable] has a
> - [probability]% probability of being [outcome] ([cutoff with units])
>
> relative to the 1991 to 2020 historical climatology.
> The probabilities may not add up to exactly 100% due to rounding.

Where cutoff uses `>` / `<` / `to` symbols instead of wordy labels:
- `> 3.5 °C` instead of "Above 3.5 °C"
- `1.8 to 3.5 °C` instead of "1.8 to 3.5 °C"
- `< 1.8 °C` instead of "Below 1.8 °C"

## Summary

- Add `<TooltipWidget />` info icon to the probability bars section title in the S2D location popup
- Tooltip shows a plain-text description of each probability bar, dynamically generated from `ProgressBarProps[]`
- Extract shared forecast category data (`buildForecastCategories`) into `s2d-build-forecast-categories.ts` — used by both the legend tooltip and the new probability tooltip
- Add `labelTooltipCutoff` field to `ProgressBarProps` for compact symbol form (`> X`, `X to Y`, `< X`) — keeps wordy `label` for bar accessibility, compact form for tooltip only
- Add JSDoc to `LocationS2DData` interface with raw-to-rendered examples from actual API data

### French tooltip — (i) icon on a separate line
When the site is in French, the probability section title is longer (e.g., "Probabilité saisonnière pour "S2D Mean Temperature":"). This causes the (i) icon to wrap to the next line. This is a visual consequence of the title length in French — not a layout bug.

### Variable name source — should we still use WordPress post titles?
The tooltip opening line uses `climateVariable.getId()` (`s2d_precip_accum` / `s2d_air_temp`) to select the correct wording, rather than `climateVariable.getTitle()` which returns the WordPress post title (e.g., "S2D Mean Temperature"). This was necessary because:
- The WordPress title includes the "S2D" prefix, which doesn't belong in user-facing tooltip text
- French grammar requires different articles and verb conjugations between "Les précipitations totales **ont**" (plural) and "La température moyenne **a**" (singular) — a single `sprintf` template can't handle both

**Question:** Should other components also move away from `getTitle()` for user-facing display names? Currently, the section heading still uses `variableName` from `getTitle()`.

### Story mutation context
Between the initial implementation and this update, the client proposed replacing the ℹ tooltip with a red "Forecast Summary" button. The team raised accessibility concerns grounded in W3C COGA (cognitive load, pattern consistency, affordance mismatch). The client responded (2026-04-04) keeping the ℹ tooltip and providing the refined bilingual template above.

## Files changed

| File | Change |
|------|--------|
| `s2d-build-forecast-categories.ts` | **New** — shared `buildForecastCategories(forecastType): DefinitionItem[]` |
| `s2d-variable-values.tsx` | Tooltip content with `labelTooltipCutoff`, opening line by variable ID |
| `progress-bar.tsx` | Added `labelTooltipCutoff` to `ProgressBarProps` interface |
| `s2d.ts` | JSDoc on `LocationS2DData` with raw-to-rendered examples |
| `map-legend-forecast-s2d.tsx` | Refactored to consume `buildForecastCategories()` |
| `fr_CA.po` / `fr_CA.mo` | Translation entries + compiled binary |

## Test plan

- [ ] EN: Open S2D location popup for Mean Temperature — verify tooltip shows correct probabilities with compact cutoffs (`> 5.3 °C`, `1.8 to 5.3 °C`, `< 1.8 °C`)
- [ ] EN: Open S2D location popup for Total Precipitation — same check with mm/day units
- [ ] FR: Same two checks — verify French translations render correctly, cutoff symbols preserved
- [ ] FR: Verify the (i) icon position is acceptable when title wraps
- [ ] Verify map legend tooltip still renders correctly (refactored to shared function)
- [ ] Verify progress bar labels unchanged (still show wordy form: "Above 5.3 °C")

## Related ticket

- CLIM-1287

### Where To Add Tooltip

*Image provided by the ticket*


renoirb:
## Screenshots 2026-03-27

**We can see the bullet list, proper spacing between lines and padding between text and surrounding box**

CLIM-1287_Snapshot_2026-03-26-21-42-04

**TooltipWidget text depending on length**

CLIM-1287_Aligning-Title-TooltipWidget-Flex_2026-03-27-18-32-32

CLIM-1287_Aligning-Title-TooltipWidget-Flex_2026-03-27-18-32-47

CLIM-1287_Aligning-Title-TooltipWidget-Flex_2026-03-27-18-35-37

apps/src/components/map-layers/s2d-variable-values.tsx
```diff
@@ -811,11 +807,79 @@ const ProbabilitiesPart = ({
...
+	) : null;
+
+	const ProbabilityHeading = !isLoaded ? (
+		<TitleLine />
+	) : (
+		<span className="flex items-start gap-1">
+			<span className="shrink">
+				<TitleLine />
+			</span>
+			<span className="shrink-0">
```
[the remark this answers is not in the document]
renoirb: Ah, I realize too late that I didn't explain things in my [earlier comment showing 3 screenshots](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/669#issuecomment-4145748811) (search for "TooltipWidget text
depending on length")

What I'm about to explain here is why it is like that, and it is a compromise because I had tried last week, and some more today.

Good catch. The gap appears specifically when the title text wraps between words — e.g. the French "Probabilit&eacute; saisonni&egrave;re pour &laquo;Temp&eacute;rature moyenne&raquo;:" is long enough to push a word to the next line. When it fits on one line (English, or shorter French variants), the icon sits tight.

That was the point of the 3 image variants with different words

<figure>
  
  Three title length variations showing icon position
  <figcaption>Three title length variations — short title (icon tight), wrapping title (gap), and stress test (long wrap).</figcaption>
</figure>

I tried several CSS approaches to close the gap when wrapping occurs, but none worked without breaking something else:

- [`flex-basis: min-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex-basis) on the title span — forces every word onto its own line:

  <figure>
    
    flex-basis: min-content result
    <figcaption><code>flex-basis: min-content</code> — text collapses to one word per line</figcaption>
  </figure>

- [`flex-basis: fit-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex-basis) — no effect, span still takes full width:

  <figure>
    
    flex-basis: fit-content result
    <figcaption><code>flex-basis: fit-content</code> — no change, span still 380px wide</figcaption>
  </figure>

- `inline-flex` on the container — same result, the `<h3>` is block-level:

  <figure>
    
    inline-flex result
    <figcaption><code>inline-flex</code> — same result, <code>&lt;h3&gt;</code> is block-level so flex item still stretches</figcaption>
  </figure>

- [`contain: inline-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/contain) — performance hint only, doesn't affect sizing

The root cause: [flex sizes boxes, not text lines](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Controlling_ratios_of_flex_items_along_the_main_axis). When text wraps inside a flex item, the item still occupies the full row width — CSS has no way to shrink the box to match where the text actually ends.

Placing the icon inline inside the text `<span>` would close the gap, but that undoes the vertical alignment the `flex items-start` wrapper was introduced to fix.

Current behavior is the best trade-off I found.

apps/src/components/map-layers/s2d-variable-values.tsx
```diff
@@ -811,11 +807,78 @@ const ProbabilitiesPart = ({
 		}
 	}
 
-	const TitleLine = sprintf(
-		frequency === FrequencyType.MONTHLY
-			? __('Monthly %s probability:')
-			: __('Seasonal %s probability:'),
-		variableName
+
+	// We use the variable id instead of the title which is more reliable.
```
ChaamC: Maybe we can move that comment closer to the relevant context (e.g. line 835? `variableId ?? 'fallback'`), since now with the updated comment, it makes a little bit less sense to be placed here in the file.

renoirb:
Screenshots

CLIM-1287_2026-04-16_Snapshot-1
CLIM-1287_2026-04-16_Snapshot-2

apps/src/lib/s2d.ts
```diff
@@ -12,16 +12,43 @@ import { __ } from '@/context/locale-provider';
...
+ *
+ * Example response (temperature, °C):
+ * ```json
+ * { "cutoff_unusually_low_p20": 5.0, "cutoff_below_normal_p33": 1.8,
+ *   "historical_median_p50": 5.6, "cutoff_above_normal_p66": 5.3,
+ *   "cutoff_unusually_high_p80": 6.9,
+ *   "prob_unusually_low": 4, "prob_below_normal": 37,
+ *   "prob_near_normal": 38, "prob_above_normal": 26,
+ *   "prob_unusually_high": 5, "skill_level": 2, "skill_CRPSS": 0.12 }
+ * ```
```
ChaamC: This feels too verbose. I think we could remove that part of the comment.

apps/src/lib/s2d.ts
```diff
@@ -12,16 +12,43 @@ import { __ } from '@/context/locale-provider';
...
+ *
+ * Cutoffs are raw values in the variable's unit (°C, mm/day).
+ * Probabilities are 0–100 percentages. Rendered as:
+ * - Bar label (wordy): "Above 5.3 °C", "1.8 to 5.3 °C", "Below 1.8 °C"
+ * - Tooltip cutoff (compact): "> 5.3 °C", "1.8 to 5.3 °C", "< 1.8 °C"
+ */
 export interface LocationS2DData {
+	/** 20th percentile cutoff, e.g. `5.0` → "< 5.0 °C" */
```
ChaamC, line 22: Same for this comment and other similar, feels like overdocumentation.

apps/src/lib/s2d.ts
```diff
@@ -12,16 +12,43 @@ import { __ } from '@/context/locale-provider';
...
+ *   "prob_unusually_low": 4, "prob_below_normal": 37,
+ *   "prob_near_normal": 38, "prob_above_normal": 26,
+ *   "prob_unusually_high": 5, "skill_level": 2, "skill_CRPSS": 0.12 }
+ * ```
+ *
+ * Cutoffs are raw values in the variable's unit (°C, mm/day).
+ * Probabilities are 0–100 percentages. Rendered as:
+ * - Bar label (wordy): "Above 5.3 °C", "1.8 to 5.3 °C", "Below 1.8 °C"
+ * - Tooltip cutoff (compact): "> 5.3 °C", "1.8 to 5.3 °C", "< 1.8 °C"
+ */
```
ChaamC: Feels out of context for this file. These are generated elsewhere. I would remove that part.

apps/src/lib/s2d.ts
```diff
@@ -12,16 +12,43 @@ import { __ } from '@/context/locale-provider';
...
+ * { "cutoff_unusually_low_p20": 5.0, "cutoff_below_normal_p33": 1.8,
+ *   "historical_median_p50": 5.6, "cutoff_above_normal_p66": 5.3,
+ *   "cutoff_unusually_high_p80": 6.9,
+ *   "prob_unusually_low": 4, "prob_below_normal": 37,
+ *   "prob_near_normal": 38, "prob_above_normal": 26,
+ *   "prob_unusually_high": 5, "skill_level": 2, "skill_CRPSS": 0.12 }
+ * ```
+ *
+ * Cutoffs are raw values in the variable's unit (°C, mm/day).
+ * Probabilities are 0–100 percentages. Rendered as:
```
ChaamC: We could keep that part if we want to document a bit. We could add that cutoffs represent values associated with specifically defined percentiles.

apps/src/components/map-layers/s2d-variable-values.tsx
ChaamC: @renoirb

apps/src/lib/s2d.ts
renoirb: Yeah. That's why I added another comment to give an idea about what it looks like. This slipped under my eyes.

apps/src/lib/s2d.ts
renoirb: I couldn't visualize what it is for and what the data might look like.

ChaamC (approved):

PR #671, opened 2026-04-02
CLIM-1223: Accurate Info in LocationModal when we either use Search or the Locate button

renoirb:
## Problem

When using the Map with interactive regions (watershed, health, census), searching for a location or using "locate me" always queries the API with `featureId=0`:

```
/get-delta-30y-regional-values/watershed/0/tx_max/ann
```

This causes every search result to show the same incorrect data, regardless of location.

## Root Cause

When a user clicks directly on the map, VectorGrid provides feature properties (including the region `id`) from the tile data automatically. The search and locate-me paths skip that tile interaction — they fire a synthetic click event with coordinates but **no layer properties**. `getFeatureId()` receives `undefined`, returns `null`, and the downstream code falls back to `0`.

## Fix

Added `fetchRegionFeatureByCoords()` in `services.ts` — a GeoServer WFS query that resolves which interactive region contains the searched coordinates. `handleLocationChange()` in `search-control.tsx` now calls this before firing the synthetic click, passing the resolved properties into the event so `handleClick()` receives the same data shape as a real map click.

- One fix covers both search and locate-me (both converge on `handleLocationChange`)
- No changes to `handleClick()` or anything downstream
- For `GRIDDED_DATA` mode, no WFS call is made (not needed)

### Files changed

| File | Change |
|------|--------|
| `services.ts` | Added `fetchRegionFeatureByCoords()` — WFS query to `CDC/ows` endpoint |
| `search-control.tsx` | `handleLocationChange()` resolves region properties before firing click |

## Testing

For each interactive region type (watershed, census, health):

1. Select a variable (e.g., Hottest Day) and the region type
2. **Click** a region shape — note the popup values and the API URL in DevTools
3. **Search** for a location within that same region — verify the popup shows the same values and the API URL uses the correct `featureId` (not `/0/`)
4. **Locate me** — verify same behavior
5. **Switch to gridded data** — verify no WFS call is made and existing behavior is unchanged

### DIFF
--- a/apps/src/components/map-layers/search-control.tsx
+++ b/apps/src/components/map-layers/search-control.tsx
@@ -15,9 +15,21 @@ import 'leaflet-search';
 import mapPinIcon from '@/assets/map-pin.svg';
 
 import { cn, parseLatLon } from '@/lib/utils';
-import { fetchLocationByCoords } from '@/services/services';
-import { SearchControlLocationItem, SearchControlResponse } from '@/types/types';
-import { LOCATION_SEARCH_ENDPOINT, SEARCH_DEFAULT_ZOOM, SEARCH_PLACEHOLDER } from '@/lib/constants';
+import { useClimateVariable } from '@/hooks/use-climate-variable';
+import {
+	fetchLocationByCoords,
+	fetchRegionFeatureByCoords,
+} from '@/services/services';
+import type {
+	SearchControlLocationItem,
+	SearchControlResponse,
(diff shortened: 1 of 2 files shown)


xfrenette-crim (approved):

ChaamC (approved):

renoirb:
PR #676 will overwrite this PR and should be about "*paying the principal*" on the TechnicalDebt.

PR #676, opened 2026-04-22
CLIM-1223: TechDebt — dispatch real DOM PointerEvent on VectorGrid (pay principal on FI1)

renoirb:
## Description

Pays principal on FI1 — the "synthetic Leaflet click" smell flagged during CLIM-1223 Pass 1. `vectorLayer.fire('click', { latlng })` is not a DOM dispatch; it's a direct procedure call with a fabricated event object, which is why the `featureId` was ever missing in the first place. This branch replaces it with a real `PointerEvent('click')` dispatched on the VectorGrid canvas tile, so VectorGrid's own hit-test resolves the feature exactly as it does for a real user click.

Fixes the same bug CLIM-1223 reported, without the WFS preflight introduced in #671, and additionally covers `recent-locations.tsx` which was on the same synthetic-click pattern.

### How it works

1. `map.setView(latlng, SEARCH_DEFAULT_ZOOM)` centers the target in the viewport.
2. Gate on `moveend` with a 500 ms timeout (`setView` fires `moveend` synchronously on short pans — `once('moveend')` alone hangs).
3. Query the grid pane's canvas elements directly (`gridPane.querySelectorAll('canvas')` + bounding-rect filter — `elementFromPoint` returns the outermost leaflet-container because Leaflet sets `pointer-events: none` on pane wrappers).
4. Dispatch `PointerEvent('click', …)` on the canvas → VectorGrid's own `_onClick` → `handleClick()` → correct `featureId` in `layer.properties`.

No synthetic fallback (broken-escalator rule): if the dispatched event hits nothing, the UI does nothing — matching what a real click at that pixel would do.

### Pass 2 implementation commits (paying principal on FI1), followed by post-wire cleanup below

1. Revert PR #671 (remove WFS preflight).
2. Strip `vectorLayer.fire('click', …)` from both call sites (intermediate state: search navigates but doesn't pop — falsifiable, bisectable).
3. Introduce `apps/src/lib/dispatch-map-click.ts` and wire `search-control.tsx` + `recent-locations.tsx`.

### Post-wire cleanup commits (ticket Session 8 — not behavior-bearing)

- `189a4dd8` — Drop dead `layerRef` prop from `SearchControl` and its `map-container.tsx` call site. Vestige of the pre-D4a vector-layer coupling; no longer needed since DOM dispatch resolves the target via `elementFromPoint` + canvas rect filter.
- `120a09fd` — Refactor `SearchControl` to arrow form with externalized `SearchControlProps`; return type `ReactElement | null` → `ReactElement` (useMap throws, never returns null); drop default `React` import. No behavior change; `tsc -b` clean.

### Verified in browser

`(search | locate-me | recent-locations) × (watershed | census | health)` — correct `featureId` in API URL, popup shows matching data. Gridded mode works via the same path (no special case).

### Why this is Prudent + Deliberate tech-debt work

Pass 1 (#671) paid *interest* on FI1 by working around the smell; this PR pays *principal* by eliminating it. Framing, preconditions, and recurrence count live in the vault TechDebt docs — see the ticket context.

## Related ticket

- CLIM-1223
- Supersedes #671 (merged, then reverted on this branch)

### DIFF
--- a/apps/src/lib/dispatch-map-click.ts
+++ b/apps/src/lib/dispatch-map-click.ts
@@ -9,7 +9,7 @@ import L from 'leaflet';
  * tiles are queried via `gridPane.querySelectorAll('canvas')` and filtered
  * by bounding rect instead.
  *
- * `Promise.race` with a 500 ms timeout guards the case where `setView`
+ * `Promise.race` with a 1,000 ms timeout guards the case where `setView`
  * fires `moveend` synchronously (short pan, no animation) before the
  * listener is attached.
  */
@@ -20,7 +20,7 @@ export const dispatchMapClick = async (
 	void latlng;
 	await Promise.race([
 		new Promise<void>((resolve) => map.once('moveend', () => resolve())),
-		new Promise<void>((resolve) => setTimeout(resolve, 500)),
+		new Promise<void>((resolve) => setTimeout(resolve, 1_000)),
 	]);
 	await new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
 	const container = map.getContainer();
--- a/apps/src/lib/dispatch-map-click.ts
+++ b/apps/src/lib/dispatch-map-click.ts
@@ -1,0 +1,55 @@
+import L from 'leaflet';
+
+/**
+ * Dispatches a real DOM PointerEvent on the VectorGrid canvas tile at the
+ * map container center after the view settles.
+ *
+ * `pointer-events: none` on Leaflet's pane wrapper divs causes
+ * `document.elementFromPoint` to return the outermost container — canvas
+ * tiles are queried via `gridPane.querySelectorAll('canvas')` and filtered
+ * by bounding rect instead.
+ *
+ * `Promise.race` with a 500 ms timeout guards the case where `setView`
+ * fires `moveend` synchronously (short pan, no animation) before the
+ * listener is attached.
+ */
+export const dispatchMapClick = async (
+	map: L.Map,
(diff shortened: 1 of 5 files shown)


apps/src/lib/dispatch-map-click.ts
```diff
@@ -0,0 +1,55 @@
...
+ * listener is attached.
+ */
+export const dispatchMapClick = async (
+	map: L.Map,
+	latlng: { lat: number; lng: number },
+): Promise<void> => {
+	void latlng;
+	await Promise.race([
+		new Promise<void>((resolve) => map.once('moveend', () => resolve())),
+		new Promise<void>((resolve) => setTimeout(resolve, 500)),
```
ChaamC: Should we increase the timeout value, to maybe 1000ms?

What is the impact of increasing this?

If the impact if a timeout that would be too low is that the popup might not appear, it could be good to increase it a bit?

Not sure, but I will rely on your judgement on this.

ChaamC (commented):
Is it just me or is there a slightly longer delay before the new popup result appears, a delay that is slightly longer than on prod, or on previous commits on the main?

I don't think it relates to the 500ms timeout. Is it just because it is a local setup?
Do you have the same behavior and do you know what could cause this?

renoirb:
It waits for `moveend`.  The time should be counted from that moment.  Unless it waits for the images tiles to wait. 

A sure thing. It's going to be longer than before.  Before it would call with anything and return bogus values.

ChaamC (approved):

PR #677, opened 2026-04-28
CLIM-1371: Fix: dayFormatter incorrectly converts day of year

xfrenette-crim:
## Description

This PR fixes 2 issues with the `dayFormatter` function. The two issues generally cancel out, but they should be fixed to prevent issues in the futur.

Also, it adds the possibility to set the first day of the year to "July 1st" instead of "January 1st".

See the description of the Jira ticket for the issue description

## Related ticket

CLIM-1371

### DIFF
--- a/apps/src/lib/format.test.ts
+++ b/apps/src/lib/format.test.ts
@@ -2,47 +2,93 @@ import { expect, test, describe } from 'vitest'
 import { doyFormatter, MonthFormat } from './format';
 
 describe('doyFormatter', () => {
-    test('returns correct first day of the year', () => {
-        const output = doyFormatter(0, 'fr-CA');
-        expect(output).toEqual('1 janvier');
+    describe('when 0 corresponds to January 1st', () => {
+        test('returns correct first day of the year', () => {
+            const output = doyFormatter(1, 'fr-CA');
+            expect(output).toEqual('1 janvier');
+        });
+
+        test('returns correct last day of the year', () => {
+            const output = doyFormatter(365, 'fr-CA');
+            expect(output).toEqual('31 décembre');
+        });
+
+        test.each([0, -20])('returns first day if lower than minimum (%s)', (value) => {
+            const output = doyFormatter(value, 'en-CA');
+            expect(output).toEqual('January 1');
+        });
+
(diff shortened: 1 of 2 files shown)


apps/src/lib/format.test.ts
```diff
@@ -2,47 +2,93 @@ import { expect, test, describe } from 'vitest'
...
-    test('outputs in French', () => {
-        const output = doyFormatter(100, 'fr-CA');
-        expect(output).toEqual('11 avril');
-    });
-    
-    test('default format is "long"', () => {
-        const output_default = doyFormatter(268, 'en-CA');
-        const output_long = doyFormatter(268, 'en-CA');
-        expect(output_default).toEqual(output_long);
+    describe('when 0 corresponds to January 1st', () => {
```
ChaamC: Should the description say `when 1 corresponds to ...` or maybe `when the 1st day corresponds to ...`, to avoid confusion about the 1st day corresponding to the value 1 and not 0?

apps/src/lib/format.ts
```diff
@@ -112,17 +112,45 @@ export const normalizePostData = async (
...
+ * @param firstDayIsJuly - If true, the first day of the year will be July 1st,
+ *        else it will be January 1st.
+ * @param monthFormat - The format to use for the month. Defaults to "long".
+ *        The values can be the same as for `Date.toLocaleDateString`.
+ */
+export const doyFormatter = (
+	value: number,
+	language: string,
+	firstDayIsJuly: boolean = false,
+	monthFormat: MonthFormat | undefined = 'long',
```
ChaamC: Do you know why we allow the undefined type for this parameter?
Is it sometimes explicitly passed as an undefined parameter?

ChaamC (approved):

apps/src/lib/format.test.ts
xfrenette-crim: Description updated

apps/src/lib/format.ts
xfrenette-crim: I removed the ` | undefined`, it was effectively unnecessary.

ChaamC (approved):

PR #681, opened 2026-04-28
CLIM-1356: Download: Add a threshold selector

xfrenette-crim:
## Description

For the Download page, add the new "Threshold" dropdown that is used for the pre-calculated `sn2mm` and `sn10mm` variables.

## Related ticket

CLIM-1356


apps/src/components/download/step-variable-options.tsx
```diff
@@ -33,27 +34,38 @@ const validateS2DVariable = (
 const validateRegularVariable = (
 	climateVariable: ClimateVariableInterface,
 ): unknown[] => {
-	const version = climateVariable.getVersion() ?? true;
-	const fields = climateVariable.getAnalysisFields() ?? [];
+	const version = climateVariable.getVersion();
+	const analysisFields = climateVariable.getAnalysisFields() ?? [];
 	const values = climateVariable.getAnalysisFieldValues() ?? {};
+	const thresholdValues = climateVariable.getThresholds() ?? [];
+	const thresholdValue = climateVariable.getThreshold();
```
ChaamC: Can be a bit confusing to have 2 variable/function name that only differs by a character, where to have different roles, one being an array of available options, and the other being the selected value.

Just commenting on this, but not sure if we want to change it (if it has too much implications on the code or if many other names have the same type of nomenclature).

ChaamC (approved):

apps/src/components/download/step-variable-options.tsx
xfrenette-crim: The variable names have been changed. But the functions `getThreshold()` and `getThresholds()` are used in many places, so changing their name would require a thorough testing of the two apps to make sure nothing broke. So I left their name unchanged.

xfrenette-crim:
I corrected a small issue in the `StepSummaryVariableOptions` component, if you want to also validate that.

ChaamC (approved):

PR #682, opened 2026-05-01
CLIM-1322 (2): Map popup name resolution — server-side rewrite + dispatchMapClick refactor

renoirb:
**Popup Title Display Discrepancy When InteractiveRegionOption.GRIDDED_DATA**

- **See:** #685

Related:
- #674 
- #682 
- #684

WIP

## Summary

Rewrites `cdc_get_location_by_coords()` and refactors `dispatchMapClick` so the map popup title and pin position match what the user asked for, across all four entry paths (autocomplete selection, raw `lat,lng` paste, locate-me, manual map click) and both languages.

## What was wrong

Per the original report and clarifications we've established back around 2026-04-17:

- **Search "Vancouver" → popup says "Mount Pleasant"** (a Vancouver neighbourhood). Same shape on Toronto → North York. Caused by the server's `cdc_get_location_by_coords()` two-pass loop: it first filtered candidate rows to `gen_term IN ('Community','Metropolitan Area')` and only fell through to a broader query when that failed. Search and click ran through the same endpoint and produced different name spaces because the click-side filter excluded rows autocomplete returned.
- **Raw `lat,lng` paste snaps to the nearest DB row**: typing `44.646445,-63.619798` (slightly West of Halifax) sent the marker to `44.647650,-63.590240` instead. Two distinct user inputs collapsed to the same response.
- **FR popup titles end with a trailing comma**: `"Sainte-Adèle, "`, `"Victoria, "`. The PHP function selected `province_fr` via column-name concatenation but never aliased the returned column back to the `province` key, so `$result['province']` was undefined in FR. `province_short` came out `null` and the `title` slot for the province abbreviation was empty.
- **Search-bar selection produces a popup at a slightly-drifted coord**: `dispatchMapClick(map, latlng)` dropped its `latlng` argument (literally `void latlng;`) and dispatched the synthetic click at the container's centre. After `setView(latlng)` tile-snapped, the centre's actual lat/lng was off the typed value by 30–120 m — observable as the Session 8 DoD-2 coord-drift hypothesis.
- **Raw-coord popups display an opaque 5-character string** (`EJK8O`, `EPYTF`, …) instead of a place name. The `locationNotFound` branch passed `locationByCoords.geo_id` to `showLocation` as the title argument; the server already returned a human-readable `title` field, but it wasn't being used.

## What this changes

Eight atomic commits on `CLIM-1322_2_Popup-Title-Name-Resolution`, one per concrete change:

1. **Alias `province_fr` back to `province`** in `cdc_get_location_by_coords()` SELECT — matches the pattern already used by `cdc_get_location_by_id()` in the same file. Fixes the trailing-comma title.
2. **Drop the `preferred_terms` two-pass; harmonise `gen_term` exclusion with `cdc_location_search()`** (`('Administrative Region','Province','Territory')`). Single nearest-neighbour query at range 0.2°. Fixes the Vancouver/Toronto and Victoria search/click name divergence.
3. **Use `locationByCoords.title` for raw-coord popup, not `geo_id`** in `apps/src/components/map-layers/search-control.tsx`'s `locationNotFound` branch. Fixes the opaque 5-char popup title.
4. **Add `Census Division` to the `gen_term` exclusion** — first attempt at the Sainte-Adèle case where the Town and the Census Division share centroid coords.
5. **Add `Census Subdivision` to the `gen_term` exclusion** — same family.
6. **Rename `dispatchMapClick` → `moveAndPointAt`; honour the typed `lat/lng`** — the function now takes responsibility for `setView` (folded in) and computes the synthetic click's `clientX`/`clientY` via `map.latLngToContainerPoint(latlng)` so the click fires at the typed coord, not the tile-snapped centre. Both call sites updated. The "dispatch click" framing was the implementation detail; the new name describes the purpose.
7. **Broaden the `gen_term` exclusion to drop all administrative-grouping rows** — adds `County`, `County Regional Municipality`, all variants of `*Municipality`, `*District`, `Region`, `Regional District`, `Subdivision`, `Unorganized Territory`, plus `Administrative Sector` (lost from the original 4-term exclusion in the previous harmonisation step). None of these are place names a user types into a search bar to find a specific spot. Resolves Sainte-Adèle.
8. **Restore the canvas-vs-`elementFromPoint` workaround comment** as an inline code comment near the canvas-enumeration block — the original JSDoc carried the explanation; it was dropped during the rename refactor and restored where it actually applies.

## Why these specific changes

Carrington's reply collapsed the (mode × entry path) ambiguity matrix — search and click must return the same names; raw `lat,lng` must keep the typed coords. The fix needed to be at the server, not in the client's title plumbing. An earlier illustration (PR #674, draft) tried passing the autocomplete title through to the popup via a `searchProvidedTitle` argument; the present PR replaces that approach. The server-side rewrite reaches the click-on-map path that no client-side title plumbing could reach, and matches behaviour across the autocomplete, raw-coord-paste, and locate-me paths because they all go through the same endpoint.

The exclusion list grew through three iterations because the data has multiple administrative-grouping rows at city centroids — Census Division (Atom 4), Census Subdivision (Atom 5), then County Regional Municipality and others (Atom 7). Each step was driven by what the data actually returned at the verification anchors. The autocomplete-only filter (Atom 2's exclusion of just `Administrative Region`/`Province`/`Territory`) was sufficient for major cities (Toronto, Vancouver, Halifax) but not for towns whose centroid coincides with their administrative parent (Sainte-Adèle).

The `dispatchMapClick` refactor was on the same path: the function existed to bridge "the user said they want to go here" with "trigger the same cascade as a manual click would", but it had been dropping the `latlng` argument and approximating the destination via the container centre. The Session 8 coord-drift was one symptom; Renoir flagged it independently this session as a "the name is a misnomer" observation about the function's purpose vs its name.

## Verification

API-level probes against `dev-en` and `dev-fr` (raw HTTP, JSON parse, no Xdebug pollution observed in any probe):

| Anchor | Pre-fix | Post-fix |
|---|---|---|
| Vancouver search-by-name | `Mount Pleasant, BC` | `Vancouver, BC` |
| Toronto search-by-name | `North York, ON` | `Toronto, ON` |
| Sainte-Adèle search-by-name | `Les Pays-d'en-Haut,` | `Sainte-Adèle, QC` |
| Victoria search-by-name (EN) | `Fairfield, BC` | `Victoria, BC` |
| Victoria search-by-name (FR) | `Victoria,` (trailing comma) | `Victoria, BC` |
| Halifax raw-coord centre | `Halifax, NS` (unchanged) | `Halifax, NS` |
| Halifax raw-coord West offset | `Halifax, NS` (snap — original client complaint) | `Chocolate Lake, NS` (no snap; `coords` carries typed coords back) |

Click-precision spot-checks (5–10 km offsets from city centroids) confirm clicks far from a city return locally-appropriate places, not the city — exactly the behaviour the wide 0.2° bounding box was a theoretical concern about. The bounding box is a SQL optimisation; `ORDER BY DISTANCE LIMIT 1` finds the truly-nearest allowed row.

## Known limitations / not done

- **Browser-level UAT pending** — the API path is verified end-to-end; the visual marker placement after Atom 6's coord-precision fix and the popup-rendering path have not been exercised in a real browser run.
- **`Urban Community` / `Railway Point` / `Railway Junction` rows are now reachable** through the click path (the original 4-term exclusion no longer applies). Empirically: a click in downtown Toronto can land on `Cabbagetown, ON` (`Urban Community`) instead of `Toronto, ON` because Cabbagetown's stored centroid is closer to the click. Within Carrington's "match autocomplete" directive, but flagged for domain-expert review on whether it's the desired behaviour for `InteractiveRegionOption.GRID` map clicks.
- **Xdebug suppression on dev for `cdc/v2` endpoints** (Slice C, separable) — not addressed. Tracked as T23.
- **EN/FR parallel-WebKit comparison workflow lift** (T25, LEARN phase) — not addressed.
- **Popup "(sometimes) doesn't open" intermittent timing** — observed during Session 8, not investigated this session. Past notes from a prior session may have a path; deferred.
- **Stakeholder direction on explicit-typing of admin-grouping names** — if a user explicitly types "Les Pays-d'en-Haut" expecting that exact popup, the broadened exclusion now returns the nearest non-grouping place at the centroid. Honouring "show what was typed verbatim" requires passing the autocomplete row's `id_code` through to a fetch-by-id, which is the architectural change descoped in the analysis plan as D1c (preserved as TechDebt FI1).

## What this does not touch

Frozen historical artefact: branch `CLIM-1322_Popup-Title-Name-Resolution` and its draft PR #674. Not amended, not pushed, not closed — preserved per Renoir's instruction.

## Documentation

- [[LLM-Context-ClimateData-Ticket-CLIM-1322#Session 9]] — work session narrative, files modified, verification matrix.
- [[LLM-Context-ClimateData-Ticket-CLIM-1322-Analysis-Plan#D4]] — filter-list shape decision (D4b broader admin-grouping list picked).
- [[LLM-Context-ClimateData-Ticket-CLIM-1322-Analysis-Plan#D5]] — moveAndPointAt refactor decision (D5b rename+fold+honour-latlng picked).
- New CIs added to the analysis plan: CI-17 (dispatchMapClick latlng-discard), CI-18 (`all_areas` schema audit), CI-19 (Sainte-Adèle admin-grouping coord-tie), CI-20 (Halifax raw-coord post-fix behaviour).

### DIFF
--- a/apps/src/components/map-layers/search-control.tsx
+++ b/apps/src/components/map-layers/search-control.tsx
@@ -95,7 +95,9 @@ const SearchControl = ({
 
 	const handleLocationChange = useCallback(
 		async (inputLatlng: SearchLatLng) => {
+			console.log('RBx 2a\thandleLocationChange useCallback inputLatlng\n', inputLatlng)
 			const latlng = convertSearchLatLng(inputLatlng);
+			console.log('RBx 3a\thandleLocationChange useCallback convertSearchLatLng(inputLatlng)\n', latlng)
 			// clear all existing markers from the map
 			map.eachLayer(layer => {
 				if (layer instanceof L.Marker) {
@@ -189,7 +191,7 @@ const SearchControl = ({
 		// @ts-expect-error: suppress leaflet typescript error
 		const searchControl = new L.Control.Search({
 			url: LOCATION_SEARCH_ENDPOINT,
-			propertyLoc: ['lat', 'lng'],
+			propertyLoc: ['lat', 'lng'], // What this is for? And Where is this used?
 			autoResize: false,
 			collapsed: false,
 			autoCollapse: false,
@@ -203,12 +205,30 @@ const SearchControl = ({
 					SearchControlLocationItem & { loc: number[]; lng: string; }
 				> = {};
 
+				console.log('RBx 0a\tL.Control.Search({ formatData })\n', response?.items );
+
+				/**
+				 * ```js
+				 * response.items = [
+				 *   {
+         *     "id": "EHHUN",
+         *     "text": "Montréal",
+         *     "term": "Ville",
+         *     "location": "Montréal; Montréal",
+         *     "province": "Québec",
+         *     "lat": "45.508822",
+         *     "lon": "-73.554077"
+         *   }
+				 * ]
+				 * ```
+				 */
+
 				response.items.forEach((item: SearchControlLocationItem) => {
 					const title = buildLocationTitle(item);
 					const loc = [parseFloat(item.lat), parseFloat(item.lon)];
 					formattedData[title] = {
 						...item,
-						lng: item.lon,
+						lng: item.lon, // Ah, here
 						loc,
 					};
 				});
@@ -228,10 +248,12 @@ const SearchControl = ({
 				}
 				// Check if the coordinates are valid if the location is empty.
 				const latLng = parseLatLon(this._input.value);
+				console.log('RBx 4a\tL.Control.Search({ locationNotFound }) will call fetchLocationByCoords\n', this._input );
 				// If the coordinates are valid, move to that location.
 				if (latLng && !latLng.isPartial) {
 					// Fetch location data
-					const locationByCoords = await fetchLocationByCoords({ lat: latLng.lat, lng: latLng.lon });
+					// /wp-json/cdc/v2/get_location_by_coords?lat=45.509234158692905&lng=-73.55346679687501
+					const locationByCoords = await fetchLocationByCoords({ lat: latLng.lat, lng: latLng.lon }); //
 					// Trigger show location.
 					this.showLocation(locationByCoords, locationByCoords.title);
 				}
@@ -249,6 +271,7 @@ const SearchControl = ({
 				}),
 			}),
 			moveToLocation: (latlng: SearchLatLng) => {
+				console.log('RBx 1a\tmoveToLocation\thandleLocationChange(latlng)\n', latlng);
 				handleLocationChange(latlng);
 			},
 		});
--- a/apps/src/components/map-layers/search-control.tsx
+++ b/apps/src/components/map-layers/search-control.tsx
@@ -21,7 +21,7 @@ import 'leaflet-search';
 import mapPinIcon from '@/assets/map-pin.svg';
 
 import { cn, parseLatLon } from '@/lib/utils';
-import { dispatchMapClick } from '@/lib/dispatch-map-click';
+import { moveAndPointAt } from '@/lib/move-and-point-at';
 import { fetchLocationByCoords } from '@/services/services';
 import {
 	type SearchControlLocationItem,
@@ -29,7 +29,6 @@ import {
 } from '@/types/types';
 import {
 	LOCATION_SEARCH_ENDPOINT,
-	SEARCH_DEFAULT_ZOOM,
 	SEARCH_PLACEHOLDER,
 } from '@/lib/constants';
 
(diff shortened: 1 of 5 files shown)


renoirb:
I'm sharing this but it's still a draft of sorts.

This should be about functionality and making sure it solves what we want to solve, and confirming our hypothesis of the improvements we've assumed such changes would give.

Once we're settled with all examples of inputs and comparing what this Change-Set does, I'll clean up any excess edits.

What I'd like to adjust if applicable:
- Ensure the SQL query is safe, we should use modern/current [prepared statements](https://www.php.net/manual/en/mysqli.quickstart.prepared-statements.php) (basically a "*dumb*" string with placeholders and native functions to properly sanitize the SQL query)

**Code clarifications:**
- The function `dispatchClick` function is meaningless to what it's actually about: move the map and show information at a specific location.
- ...

PR #686, opened 2026-05-19
CLIM-1367: S2D Location Popup Forecast Summary

renoirb:
## Description

Adds a **"Summary" / "Résumé"** pill button to the S2D forecast pop-up (LocationModal). It opens a popover with a copy-paste-friendly paragraph summarising the forecast for the currently-selected location — in EN and FR, for both **Expected** ("above / near / below normal") and **Unusual** ("unusually high / low") forecast types.

Post–Story-Mutation continuation of CLIM-1287: copies CLIM-1287's bar-chart tooltip text into a new, richer popover surface and adds the lines the client's Word document asked for.

## What the popover renders

- **Title** — `Forecast Summary for {location}:`
- **Opening line** — `The {time period} {variable} has a …`. Variable-aware templates per `variableId` (`s2d_air_temp`, `s2d_precip_accum`, + fallback): French grammar can't share one `sprintf` template — "Les précipitations totales … sont de" (plural) vs. "La température moyenne … a" (singular). Only the time-period slot is interpolated; the variable label is hard-coded per variant.
- **Probability bullets** + "relative to the 1991 to 2020 historical climatology."
- **Skill-level sentence** — `The skill level is "{label}". {guidance} for this location, time period, and month of release.`
- **Inconclusive-outcome caveat** — prepended to the historical-median line when every probability bar is below threshold (< 40% Expected, < 30% Unusual).
- **Historical-median line.**
- **Release date** — via the existing `<S2DReleaseDate />`.
- **Closing line** — "Consider checking back for updated forecasts!"

## Key code changes

- **New** `apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx` — the component; each line is its own small `<Line… />`.
- **New reusable primitive** `apps/src/components/ui/modal-summary-popover.tsx` — brand-blue pill `Button` + Radix `Popover`, generic over children / side / trigger node / title. First consumer is this button; future "summary in a popover" surfaces can reuse it.
- **Currently-selected location is now first-class** in `apps/src/features/map/map-slice.ts`: added `selectedLocation` as the source of truth, a `setSelectedLocation` action, and a memoised `selectSelectedLocationTitle`. The old `recentLocations[last]` read was unreliable — `addRecentLocation` de-dupes by id, so re-clicking a seen location is a no-op and the last entry isn't the current selection.
- **Shared S2D helper** `generatePeriodRangeLabel(dateRangeStart, frequency, locale)` lifted into `apps/src/lib/s2d.ts` so the bar-chart caption and this component render identical period labels.
- **`apps/src/lib/format.ts`** — new `formatIntlDate(dateLike, locale, options?)` (wraps `Intl.DateTimeFormat`, `formatUTCDate` fallback); `formatPeriodRange` gains a `dateFormat` param.
- **i18n** — EN msgids in code + FR in `fw-child/languages/react-apps/fr_CA.po` (+ compiled `.mo`).

## Notes for reviewers / future work

- The per-variable whole-sentence msgid pattern works around a `@wordpress/i18n` limit (gettext can't pick a translation form by the gender/number of a substituted value). A 3rd S2D variable falls through to an EN-correct / FR-degraded `sprintf` fallback until given its own msgid pair.
- `<ModalSummaryPopover />` is intentionally generic in `components/ui/` despite S2D being its only current consumer.

## Related ticket

[CLIM-1367](https://ccdpwiki.atlassian.net/browse/CLIM-1367)

[CLIM-1367]: https://ccdpwiki.atlassian.net/browse/CLIM-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+		const title = temporaryHack && temporaryHack.title ? temporaryHack.title : '';
+		forecastSummaryFirstLine = (
+			<p className="mt-2 font-semibold">
+				{sprintf(
+					__('Forecast Summary for %s:'),
+					title,
+				)}
+			</p>
+		);
+	}
```
renoirb: This is because we can't pass the value of the current popup because it's going through `climateVariable?.getLocationModalContent({})` and there's no way we can pass the title all around. So we have to poke a hole with `useAppSelector`.

image

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+						1,
+						locale
+					),
+				)}
+			</MaybePrefixWhen>
+			<p className="mt-2">
+				{__('The probabilities may not add exactly to 100% due to rounding.')}
+			</p>
+			<p className="mt-2">
+				<S2DReleaseDate className="-font-semibold -uppercase" tooltip={false} />
```
renoirb: `S2DReleaseDate` is well made, another place we can re-use it! \o/

renoirb:
Screenshot from 2026-05-22 01-06-57

fw-child/languages/react-apps/fr_CA.po
```diff
@@ -1240,21 +1240,48 @@ msgstr "L'étendue près de la normale est définie à l'aide de la climatologie
...
 
 msgid  "Monthly %s probability:"
 msgstr "Probabilité mensuelle pour \"%s\":"
 
 msgid  "The total precipitation has a"
-msgstr "La précipitation totale a"
+msgstr "Les précipitations totales sont de"
+
+msgid  "The %s total precipitation has a"
+msgstr "Les précipitations totales pour %s sont de"
```
ChaamC: Change "sont de" for "ont"

fw-child/languages/react-apps/fr_CA.po
```diff
@@ -1240,21 +1240,48 @@ msgstr "L'étendue près de la normale est définie à l'aide de la climatologie
...
+
 msgid  "Seasonal %s probability:"
 msgstr "Probabilité saisonnière pour \"%s\":"
 
 msgid  "Monthly %s probability:"
 msgstr "Probabilité mensuelle pour \"%s\":"
 
 msgid  "The total precipitation has a"
-msgstr "La précipitation totale a"
+msgstr "Les précipitations totales sont de"
```
ChaamC: Change "sont de" for "ont"

apps/src/lib/format.ts
```diff
@@ -253,3 +260,55 @@ export function formatValue(value: number, unit: string | undefined, decimals: n
...
+ * Format a date into a localized string using Intl.DateTimeFormat, with error handling.
+ *
+ * @param maybeDate - Date or string
+ * @param locale - Locale to use for formatting
+ * @param options - Intl.DateTimeFormatOptions for formatting the dates
+ *
+ * @returns A formatted date string, e.g. "2025-10-15"
+ */
+export const formatIntlDate = (
+	maybeDate: Date | string,
```
ChaamC: Not sure about the usage of "maybe" in variable names, even though I understand it is because there are 2 possible types here.
I would suggest using a name like "dateInput" instead.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+	 */
+	preamble: string;
+};
+
+const OUTCOME_PROBABILITIES_PREAMBLE = new Map<
+	ForecastType,
+	MaybePrefixWhenAllAreLessThan
+>([
+	[
+		// CLIM-1367 requirements: For Expected Conditions, (...) is shown if all probabilities are less than 40%:
```
ChaamC: I don't think we need to reference the Jira task explicitly here, we could remove those comments.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+	progressBars: ProgressBarProps[];
+};
+
+type WithChildren = {
+	children?: React.ReactNode[] | React.ReactNode;
+};
+
+type MaybePrefixWhenProps = ForecastTypeAndProgressBars & WithChildren;
+
+type MaybePrefixWhenAllAreLessThan = {
```
ChaamC: Not sure about those names with the "maybe" prefix. 
Maybe it could be renamed to something like `PreambleWhenBelowThresholdProps`, or `ConditionalPreableProps`, but I'll let you decide if you want to review those names or not.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+	progressBars,
+	children,
+}: MaybePrefixWhenProps): React.ReactNode => {
+	const currentCondition = OUTCOME_PROBABILITIES_PREAMBLE.get(forecastType);
+	if (!currentCondition) {
+		return null;
+	}
+
+	const childNodes =
+		children && Array.isArray(children) ? children : [children]; // I miss Vue.
```
ChaamC: Remove comment.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+	// CLIM-1367 requirements: (...) [shown when] ALL probabilities are less than (...)
```
ChaamC: Remove comment mentioning specific Jira task

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+	);
+	let historicalMedianPreamble = '';
+	if (allAreSmallerThan.includes(false)) {
+		// In other words; not "all percents smaller than threshold"
+		historicalMedianPreamble = '';
+	} else {
+		if (forecastType === ForecastTypes.EXPECTED) {
+			historicalMedianPreamble = currentCondition.preamble;
+		} else if (forecastType === ForecastTypes.UNUSUAL) {
+			historicalMedianPreamble = currentCondition.preamble;
```
ChaamC: Since, both if and else if do the same, couldn't they be merged into a single 'OR' condition?

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+			formattedPeriodRange = periodRange[0];
+		}
+	}
+
+	// A single sprintf template can't produce correct French for both variables:
+	// "Les précipitations totales ont" (plural) vs "La température moyenne a" (singular).
+	const tooltipOpeningLineVariants = {
+		s2d_precip_accum: sprintf(__('The %s total precipitation has a'), formattedPeriodRange) /* from climate-variables.config.ts */,
+		s2d_air_temp: sprintf(__('The %s mean temperature has a'), formattedPeriodRange)        /* from climate-variables.config.ts */,
+		fallback: sprintf(__('The %s %s has a'), variableName.toLowerCase(), formattedPeriodRange), // THIS WON'T WORK UNLESS WE USE POSITIONAL -- TODO Check
```
ChaamC: Can we address this TODO?
Or else, comment/update this line and the comment?

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+	`uppercase`,
+] as const;
+
+export type ForecastSummaryPopoverProps = ForecastSummaryContentsProps & {
+	className?: string;
+	popoverContentSide?: PopoverContentSide;
+};
+
+/**
+ * @TODO NOTE FOR DISCUSSION:
```
ChaamC: Can we address this TODO? I'd rather not leave TODOs in the code.

It is acceptable to me that we have a unique object for the Forecast Summary, since it is a bit different than others, visually.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
ChaamC: For the period range, is it possible to use the long form of the periods?

E.g. : use 'May to July' instead of 'May to Jul'

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+	// Category definitions parallel to progressBars, for tooltip content
+	const forecastCategories = buildForecastCategories(forecastType);
+
+	const CN_RELEASE_DATE = [
+		'-font-semibold',
+		'-text-dark-purple',
+		'-text-xs',
+		'-tracking-wider',
+		'-uppercase',
+	];
```
ChaamC: The release date appears slightly different visually than other texts in the popup, is it possible to fix it?

I am not sure, but it might just be the spacing in between the letters that needs to be increased.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,365 @@
...
+						variant="outline"
+						className={cn(
+							'mr-1',
+							CN_BUTTON_EFFECTS,
+							CN_ROUNDED_BIG_BUTTON_TEXT,
+						)}
+						title={__('Forecast Summary')}
+					>
+						<BookOpenText size={16} />
+					</Button>
```
ChaamC: Can we use the visuals proposed in the document?

Image

I am seeing this :

Image

I would prefer if we use what they suggested.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
ChaamC: There is a bug with the time period displayed.
It doesn't always display the seasonal or monthly version of the time correctly.

With the forecast type "Expected Conditions", I always see the seasonal format of the time period (even if I select the Monthly frequency).

With the forecast type "Unusual Conditions", I always see the monthly format of the time period (even if I select the Seasonal frequency).

ChaamC (commented):
Tested, and works well in general, but some comments to address.

apps/src/lib/format.ts
renoirb: > (...) even though I understand it is because there are 2 possible types here. 

Exactly. And *maybe* it's of `Date` type.

> I would suggest using a name like "dateInput" instead.

... isn't everything in a web app... probably an *input*? :)

The thing here is that sometimes, we have a `Date` to format... and some other times, we have a `string` that looks like a date to format.  And that function input argument. Is an input. So would we call this an `dateInputInput`? :)

But, we're accepting what is a considered as a Date (object or string).

Maybe; `eitherDateStringOrObj`

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: Right. No problem.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: > Not sure about those names with the "maybe" prefix.

Yeah. It's a component that treats the possibility of having something with a prefix. I'm unsure about that name use here either.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: When I wrote this block, I was thiking about using exactly the same as `TooltipWidget` but with differences.

But since it's different than a tooltip. It's like a "please open a panel with a text representation of what's inside this box the button is on" kinda thing.

We could rename it like that later. Or I'd make that `ForecastSummaryPopover` in its own file and name it like the intent this is for: ... a button that opens and displays a more complex text contents with HTML and lots of conditionals... like what we're doing.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: Have a look at [`DateTimeFormat` constructor options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#date-time_component_options)

For Month:
- ["numeric"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#numeric)
E.g., 3

- ["2-digit"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#2-digit)
E.g., 03

- ["long"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#long_3)
E.g., March

- ["short"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#short_3)
E.g., Mar

- ["narrow"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#narrow_3)
E.g., M. Two months may have the same narrow style for some locales (e.g., both March's and May's narrow styles are M in the en-US locale).


And, it's very well informed about language specifics. So, in french it'll be shortening longer words properly. Differently. Because it's a mess typical translation conventions.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: Ah. Of course. That conditional was where I was doing before, then had the concept of `currentcondition` lifted up above... and mindlessly left this there.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: That's the fallback so it shows something (instead of an error). We can remove the comment though.

apps/src/lib/format.ts
ChaamC: I'll let you judge if you want to change it or not, it was just a suggestion.

I just thought it was more clear.

Not sure about adding the type in the name either, but maybe something like `rawDate` could do too, to indicate it might not be formatted yet to the right format.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
ChaamC: Why not use the 'long' format, as it is what is used in the example they gave us, and it is used in the main popup too?

apps/src/lib/format.ts
renoirb: > (...) maybe something like rawDate could do too

Yes. But. It's not a "raw" date. It's either a Date object. OR a string. And BOTH are accepted.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: I would prefer too.  But spacing breaks as soon as le texte en Français entre en jeu.

image

OR.

I also make the text of the button to wrap. I tried. I couldn't do the normal way of saying a width.  It could be a ratio with the date.  Didn't work well with current architecture.

As I said. I tried.

Alternatively. I could use a dumb. Fresh from the 1980s... `<BR>`. `<blink><marquee>what do you think</marquee></blink>`

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: Notice the requirements and the difference for displaying date range when is *Unusual Conditions* (or the other).  One of them has a difference to use just the first date, or the second.  (I'm not in front of the code while writing this)

And, I'll check more closely what you're saying.

renoirb: long is now used.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: ah. uhm. maybe there's another className that I didn't pick.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: >  I always see the monthly format of the time period

You mean only one month.

Yeah, it's because I wasn't checking with `frequencyType === S2DFrequencyTypes.SEASONAL` like I just changed.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -0,0 +1,443 @@
...
+			formattedPeriodRange = timePeriodTuple[0];
+		}
+		if (
+			timePeriodTuple.length === 2 &&
+			timePeriodTuple[0] === timePeriodTuple[1]
+		) {
+			// Fail-Safe when we have two times the same month, so we don't display 'may to may'.
+			formattedPeriodRange = timePeriodTuple[0];
+		}
+	}
```
renoirb, line 233: This is a follow-up from [this comment](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/686#discussion_r3307285772) about displaying only one month or two.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -84,77 +82,201 @@ const OUTCOME_PROBABILITIES_PREAMBLE = new Map<
...
+	if (currentConditions) {
+		const allPercent = progressBars.map(({ percent }) => percent);
+		const allPercentLessThan = allPercent.map(
+			(percent) => Math.floor(percent) < currentConditions.threshold
+		);
+		if (allPercentLessThan.includes(false)) {
+			// In other words; NOT "all percents smaller than threshold"
+		} else {
+			parts.push(currentConditions.preamble);
 		}
```
ChaamC: We could improve this code, to juste have one 'if' with the inverted condition.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
```diff
@@ -285,81 +402,42 @@ export const ForecastSummaryContents = (
...
-		popoverContentSide = 'top',
 	} = props;
 
-	return (
+	// 'Forecast Summary' in French would be 'Résumé des Prévisions'
+	const popoverTriggerButtonTitle = __('Forecast Summary');
+	// The button text, when is the same as 'Forecast Summary' when
+	// in french, the Release Date contents may wrap text in a wrong
+	// way so we have to keep this button small.
+	// We could alternatively use BookOpenText from 'lucide-react'
```
ChaamC: Update this comment, since we have opted for another solution now.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
ChaamC: @renoirb Still seems a bit different, no?

I think it might be the spacing in between the letters?

For example, if I compare the word "de" from other text in the summary :
image

VS the word "de" in "Date de sortie" :
image

ChaamC (commented):
Ok, ça marche au niveau fonctionnalité.

Juste 2 petits commentaires restants que je voulais tu corriges (un commentaire à retirer/adapter, pis un `if` que je suggérais de restructurer).

Enlever aussi les `console.log`, comme tu me l'avais mentionné.

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: Fixed. The wording needed to be improved.

```ts
const allPercent = progressBars.map(({ percent }) => percent);
const eachPercentLessThan = allPercent.map(
    (percent) => percent < rule.threshold
);
const forecastHasNoClearOutcome = eachPercentLessThan.every(
    (isLessThan) => isLessThan === true
);
if (forecastHasNoClearOutcome) {
    parts.push(rule.caveatText);
}
```

apps/src/components/map-layers/s2d-location-modal-forecast-summary.tsx
renoirb: Remplacé par:

```
// If we used the same as `popoverTriggerButtonTitle` we'd see
// S2DReleaseDate's date wrapped into this button on long-month dates.
```

ChaamC (approved):
Good job! All good now

PR #687, opened 2026-05-22
CLIM-1395: Fix: Download: AHCCD blocked at step 3

xfrenette-crim:
## Description

Issue: in Download, when selecting a "AHCCD" variable, the user cannot go pass the step 3 because the "next step" button is disabled.

The reason is because the code validates that a version has been selected, but the AHCCD variables don't have versions.

## Related ticket

CLIM-1395


ChaamC (approved):
All good!

PR #690, opened 2026-05-29
CLIM-1367 (2): S2D ProgressBar rounding to be consistent

renoirb:
## Description

The rounding of value to consider if a probability is within or outside of range has to be the same.

### Report

The fact that we are at a percent that shows rounded to a cutoff (here being `40%`), and is writing that it isn't is confusing.

|  |  |
|--|--|
| CLIM-1367_Pre-2-Feedback_01 | CLIM-1367_Pre-2-Feedback_02 |

## Related ticket

- #686

### DIFF
--- a/apps/src/components/map-layers/s2d-variable-values.tsx
+++ b/apps/src/components/map-layers/s2d-variable-values.tsx
@@ -14,6 +14,7 @@ import { cn, findCeilingIndex, utc } from '@/lib/utils';
 import {
 	extractSkillLevelData,
 	generatePeriodRangeLabel,
+	normalizeProbabilitiesBarChartPercent,
 	type LocationS2DData,
 } from '@/lib/s2d';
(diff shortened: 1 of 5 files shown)


apps/src/components/map-layers/s2d-variable-values.tsx
```diff
@@ -187,6 +190,7 @@ export const getProbabilityColour = (
 	const colours = colorMap.colours as `#${string}`[];
 	const defaultColor = '#909090';
 	let percentageForQuery = Math.round(percentage);
```
ChaamC: Il faudrait juste supprimer cette ligne, car sinon la variable `percentageForQuery` est déclarée deux fois avec la ligne suivante.

renoirb: ah!? ok.

ChaamC (approved):

PR #691, opened 2026-06-02
CLIM-1096: Map Sidebar EmissionScenarios Default option SSP2-4.5 (RCP4.5) when applicable

renoirb:
## Description

On the **Map**, the default emission scenario now resolves to **SSP2-4.5** (CMIP6 `ssp245`) / **RCP4.5** (CMIP5 `rcp45`) uniformly across all variables, instead of each variable's first-listed scenario.

It is resolved at the model layer: a new `getDefaultScenario()` on `ClimateVariableBase`, backed by a shared `resolveDefaultScenario()` cascade that both `getScenario()`'s fallback and `getValidScenarioForVersion()` delegate to (single source of truth, no UI duplication).

**Version-aware cascade**

1. Prefer the bare preferred scenario — `ssp245` (CMIP6) / `rcp45` (CMIP5)
2. Else the `-p50` median — e.g. `rcp45-p50` for percentile-only variables like `sea_level`
3. Else the first available scenario; empty → `null`

**Scope & safety**

- Map-only in behaviour — Download / Analyze use the multi-select `analyzeScenarios` path and S2D has no scenarios, so they are unchanged.
- An explicit `?scen=` selection still wins; only the no-selection default changes.
- New `climate-variable-base.test.ts` covers CMIP5/CMIP6 × bare/percentile, the explicit-selection guard, and empty-scenarios.

> [!NOTE]
> Acceptance is still pending the ticket creator's confirmation on scope (Map-only vs global) and the meaning of "always". The mechanism is independent of those answers; the percentile default (`rcp45-p50` median) is already customer-confirmed.

## Related ticket

- CLIM-1096

### DIFF
--- a/apps/src/components/download/step-additional-details.tsx
+++ b/apps/src/components/download/step-additional-details.tsx
@@ -500,7 +500,7 @@ const StepAdditionalDetails = React.forwardRef<
 					<div className="mb-8 max-w-md">
 						<CheckboxFactory
 							name="emission-scenarios"
-							title={__('Emissions Scenarios') + ' TESTING FOR CLIM-1096: This is from download side'}
+							title={__('Emissions Scenarios')}
 							tooltip={<EmissionScenariosTooltip />}
 							orientation="horizontal"
 							className="max-w-md"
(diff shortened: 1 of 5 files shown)


ChaamC (approved):
All good, tested and works well!
Good job and thanks for adding related tests!

PR #693, opened 2026-06-08
CLIM-1393: Time series: Make 30-year selected by default + other fixes

xfrenette-crim:
**Context:** currently, in the Maps page, when opening the time series graph, the default selected tab is the "Annual values".

**Problem:** The tab "30-year averages" should be the one selected by default.

**This PR**

This PR makes the default tab to be "30-year averages".

But this PR also uses the chance to improve a little bit the code of the time series graph panel:

* The concept "annual values" is renamed "individual values" which better reflects what this tab is actually doing (but the tab's text is not changed).
* For the same reason, the concept "30 year values" is renamed to "period values", removing possible confusion if we ever get a period of other than 30 years (but the tab's text has not been changed).
* The default tab is "30-year averages", but there is now a fallback if it's not enabled/available.
* A small issue was fixed: some variables don't have data for all the series (ex: Humidex doesn't have SSP3-7.0), but the series' name is still shown in the legend. The issue is fixed by removing series that don't have any data.

## Related issue

CLIM-1393

### DIFF
--- a/apps/src/components/climate-data-chart.tsx
+++ b/apps/src/components/climate-data-chart.tsx
@@ -465,6 +465,8 @@ const ClimateDataChart: React.FC<{
 			}) || []
 		);
 
+		// Remove series without any data (else they would be shown in the
+		// legend but not in the chart).
 		return series.filter((s) => (s.data?.length ?? 0) > 0);
 	}, [activeTab, activeSeries, seriesObject]);
 
--- a/apps/src/components/climate-data-chart.tsx
+++ b/apps/src/components/climate-data-chart.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import Highcharts, {
 	numberFormat,
 	Options,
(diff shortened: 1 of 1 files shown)


apps/src/components/climate-data-chart.tsx
```diff
@@ -26,10 +26,21 @@ import { useClimateVariable } from "@/hooks/use-climate-variable";
...
+ * impacts only the value (i.e. popup) shown when hovering the graph, it doesn't
+ * impact the graph itself.
+ *
+ * - 'individual-values': The value of each individual point is shown.
+ * - 'period-averages': The average of the values of the period is shown (ex:
+ *     average over a 30-year period).
+ * - 'period-changes': The _difference_ between the period average (like
+ *     'period-averages') and a reference period.
+ */
+type TabValue = 'individual-values' | 'period-averages' | 'period-changes';
```
ChaamC: Would it be relevant to store these 3 values as 3 constants, since they are used multiple places in this file?

ChaamC (approved):
LGTM, just a minor suggestion. I like the changes to the names. 
Tested and seems to work well too, good job!

PR #694, opened 2026-06-08
CLIM-1364 Add Return Period variables

xfrenette-crim:
## Description

Add the new Return Period variables and update the Maps and Download UI to support them:

* Maps
  * Fix an issue when the name of the variable is too long in the page's header 
* Maps and Download
  * Change the label "Threshold values" for "Return periods"
  * Make the values translatable
* Maps' time series graph
  * Hide the "Annual values" tab
  * Change the tab label "30-year averages" to "30-year values"
  * Extend the x-axis above the values to make sure the 30-year gray block can be shown completely

## Related ticket

CLIM-1364


fw-child/languages/react-apps/fr_CA.po
```diff
@@ -1584,3 +1584,30 @@ msgstr "REMARQUE: Le propriétaire des données nous a informés de la présence
...
+msgstr "1 sur 30"
+
+msgid "1-in-50"
+msgstr "1 sur 50"
+
+msgid "30-year values"
+msgstr "Valeurs sur 30 ans"
+
+msgid "Return Periods"
+msgstr "Niveaux de retour"
```
ChaamC: Would be "Périodes de retour"

fw-child/languages/react-apps/fr_CA.po
```diff
@@ -1584,3 +1584,30 @@ msgstr "REMARQUE: Le propriétaire des données nous a informés de la présence
...
+
+msgid "Return Period"
+msgstr "Niveau de retour"
```
ChaamC: Would be "Période de retour"

apps/src/components/sidebar-menu-items/threshold-values-dropdown.tsx
```diff
@@ -15,6 +15,9 @@ import { useClimateVariable } from "@/hooks/use-climate-variable";
 
 const ThresholdValuesDropdown: React.FC = () => {
 	const { climateVariable, setThreshold } = useClimateVariable();
+	const isReturnPeriod = climateVariable?.getClass() === 'ReturnPeriodClimateVariable';
+	let options = climateVariable?.getThresholds() ?? [];
```
ChaamC, line 19: Do we want to refactor, so that the `thresholds` references becomes more generic, since it now can contains either thresholds or return period values?

Not sure how much work this it though, but it might be cleaner on the long term.

ChaamC (commented):
Minor comments, good work!

apps/src/components/sidebar-menu-items/threshold-values-dropdown.tsx
xfrenette-crim: The term "threshold" is used in many places (~70 places), so it might be a little project in itself renaming and testing it. I suggest we don't rename it for now.

ChaamC (approved):

PR #695, opened 2026-06-08
CLIM-1394: S2D Map Skill Tooltip text use 'hatching' instead of 'cross-hatching' in text

renoirb:
## Related ticket

- CLIM-1394


ChaamC (commented):
Seems all good, but don't we need to also push the updated fr_CA.mo file?

ChaamC (approved):
Super, I like the addition on the dev.sh script too!
We can merge.

PR #696, opened 2026-06-10
CLIM-1462: Download Step reset-payload management refactor (split out of CLIM-1410)

renoirb:
## Description

**This PR is ticket CLIM-1462 — the Download SPA's step reset-payload management refactor — not the browser-navigation feature.**

This work began under CLIM-1410 (make the browser **Back** button step within the Download SPA instead of exiting it).
Analysis showed the SPA's per-step reset was driven by `useImperativeHandle` ref-walking — work a derivation should own — and that the navigation feature could not be built cleanly on top of it.
The work was therefore split, and this structural pass was given its own ticket:

- **This PR → CLIM-1462.** Structural step reset-payload management refactor only, behaviour-neutral versus production. Created as its own ticket because the refactor must land in `main` regardless of how the Back/Forward UX is ultimately specified (and it can serve CLIM-1412).
- **PR #704 → stays CLIM-1410.** The browser Back/Forward implementation (History API wiring, step-tag, history reconciliation), preserved on `CLIM-1410_2_Download-Back-Button-Navigate` for when the navigation UX requirements are gathered.

### What's included

- Mount-independent step reset-payload derivation in `apps/src/lib/download/` (`buildResetPayloadForStepsAfter`, `determineStepApplicable`) with unit tests — the reset payload a step contributes when the user navigates back past it is now derived from state.
- [Imperative-handle](https://react.dev/reference/react/useImperativeHandle) plumbing removed; the six step components are now plain function components.
- Step reset side-effects relocated into the provider as applicability-gated dispatches.
- `resolveStepNumberInFullFlow` and its test suite: the SPA counts steps in two distinct spaces — the number the user sees on screen (the stepper renumbers what remains when a variable hides steps) and the step's rank in the "full flow". The stepper shows its own sequential numbering, while reset gating compares full-flow numbers — so on-screen numbers must be translated at that boundary.

During work, we've found and had to fix a regression when editing an earlier step of a station variable (a specific type of `climateVariable`) that wiped the very step being navigated to. The fix names both numbering spaces and adds tests so the distinction stays explicit and visible.

### What's deferred

The browser Back/Forward step-navigation feature itself — CLIM-1410, via PR #704. Tracked follow-ups (re-validation on Forward, provider-level tests, the double-advance defect) travel with it, out of scope here.

### Behaviour

No user-visible change versus production: browser Back on `/download/` still exits the SPA exactly as it does today. This pass is internal structure only.

## Related tickets

- CLIM-1462 — this PR (structural refactor)
- CLIM-1410 — parent work; browser Back/Forward feature, continued in PR #704
- CLIM-1412 — may build on this refactor (noted in the CLIM-1462 ticket)


apps/src/components/download/step-additional-details.tsx
```diff
@@ -631,9 +594,9 @@ const StepAdditionalDetails = React.forwardRef<
 			}
 		</StepContainer>
 	);
-});
+};
 
-StepAdditionalDetails.displayName = 'StepAdditionalDetails';
+StepAdditionalDetails.displayName = 'StepAdditionalDetails'; // Explicit string literal, or this name would be lost in production.
```
ChaamC, line 599: For my personal comprehension, can you tell me more about this comment and why the name would be lost in production?

apps/src/context/download-provider.tsx
```diff
@@ -4,84 +4,85 @@
...
 	// Initialize URL sync
 	useDownloadUrlSync();
 
 	const [steps, setSteps] = useState<typeof STEPS>([...STEPS]);
 	// Start at step 2 if URL has variable parameter
 	const params = new URLSearchParams(window.location.search);
 	const hasVariable = params.has('var');
-	const [currentStep, setCurrentStepLocal] = useState<number>(hasVariable ? 2 : 1);
+	const [currentStep, setCurrentStepLocal] = useState<number>(
+		hasVariable ? 2 : 1
```
ChaamC: I am wondering if this should use the `DOWNLOAD_STEPS` enum values instead of `2`and `1`?

apps/src/context/download-provider.tsx
```diff
@@ -95,84 +96,89 @@ export const DownloadProvider: React.FC<{ children: React.ReactNode }> = ({
 				return [...STEPS];
 			}
 
-			if (climateVariable?.getClass() === 'StationClimateVariable' || climateVariable?.getClass() === 'StationDataClimateVariable') {
+			if (
+				climateVariable?.getClass() === 'StationClimateVariable' ||
+				climateVariable?.getClass() === 'StationDataClimateVariable'
+			) {
 				// skip step 3 (variable options) if it's a station variable
 				const skipIndexes = [2];
```
ChaamC: I am wondering if this should use the `DOWNLOAD_STEPS` enum values instead of `2`?
Same thing for following references `4` and `5`. Although I notice they seem to be 0-based instead of 1-based.

apps/src/lib/download/build-reset-payload-for-steps-after.ts
```diff
@@ -0,0 +1,202 @@
...
+	type ResetPayload,
+	type StepResetAccumulator,
+} from './types';
+
+/**
+ * Build the reset-payload contribution for step 3 (Variable Options).
+ *
+ * @remarks
+ * Mirrors the `getResetPayload` imperative handle in
+ * `components/download/step-variable-options.tsx` field-for-field. The
```
ChaamC: I don't think we should refer to some code that will not exist after merging this PR (aka the `getResetPayload` from `step-variable-options.tsx`.

This commend applies to other similar comments in the following functions.

apps/src/lib/download/build-reset-payload-for-steps-after.ts
```diff
@@ -0,0 +1,202 @@
...
+ * @remarks
+ * This is the mount-independent replacement for the former `reduce` over
+ * mounted step refs in `resetStepsAfter` (`context/download-provider.tsx`). It
+ * walks the applicable steps in ascending ordinal order and shallow-merges each
+ * step's contribution, so for keys shared between steps (`frequency` and
+ * `averagingType` appear in both step 3 and step 5) the later step wins —
+ * exactly matching the legacy ascending `reduce` merge order.
+ *
+ * Shape-table provenance: the per-step field sets and their gating come from
+ * the verified Path A plan, cross-checked against the three step components
```
ChaamC: Is the Path A referencered here the Option A in the various plans you proposed as solutions for the feature implementation? 
If so, this shouldn't be referenced in the comment here.

apps/src/lib/download/types.ts
```diff
@@ -0,0 +1,51 @@
...
+ * constants used in its place.
+ */
+export const DOWNLOAD_STEPS = {
+	additionalDetails: 5,
+	dataset: 1,
+	location: 4,
+	result: 7,
+	sendRequest: 6,
+	variable: 2,
+	variableOptions: 3,
```
ChaamC: Would it be more logical to order these by the value they have? (ordering them in the order of numbers, to reflect the order of the steps on the site)

apps/src/context/download-provider.tsx
```diff
@@ -95,84 +96,89 @@ export const DownloadProvider: React.FC<{ children: React.ReactNode }> = ({
...
 				// skip step 5 (additional details) if it's a station variable (but not station variable)
```
ChaamC: Not from this PR, but I am not sure I understand this comment, especially the part `if it's a station variable (but not station variable)`. Do you know what they meant here and if it should be rephrased?
Should it be mentioning station data instead of station variable for the parenthesis comment?

apps/src/lib/download/determine-applicable-steps.ts
```diff
@@ -0,0 +1,89 @@
+import type { ClimateVariableInterface } from '@/types/climate-variable-interface';
+import { DOWNLOAD_STEPS } from './types';
+
+/**
+ * Class names whose variables are treated as "station" variables by the
+ * Download wizard's step-skipping logic.
+ *
+ * @remarks
+ * Source of truth: the steps `useEffect` in `download-provider.tsx` (the
+ * `getClass()` comparison that gates step skipping). Kept as a private `Set`
```
ChaamC: Instead of mentioning the other file in a comment, shouldn't we fix `download-provider.tsx` to use this new Set constant? I think it would be cleaner

apps/src/lib/download/determine-applicable-steps.ts
```diff
@@ -0,0 +1,89 @@
...
+]);
+
+/**
+ * Climate-variable ids for which step 6 (Send Request) is skipped because there
+ * is no file format to choose.
+ *
+ * @remarks
+ * Source of truth: the steps `useEffect` in `download-provider.tsx`. These are
+ * the "Future Building Design Value Summaries" and "Short-duration Rainfall IDF
+ * Data" datasets.
```
ChaamC: Here again, instead of mentioning the other file, shouldn't we update the `download-provider.tsx` file to use this new Set ?

ChaamC (changes requested):
I tested it on a local setup, and I found some differences with prod.

For Climate Normals / AHCCD / Future Building Design Values / IDF Rainfall Data / Station Observations :
- Arriving at step 4, then clicking Pencil icon to modify Location step → locations are cleared up
- Same for Station Observations for date values if going back to Step 4 from Step 5

On production, those values are conserved in those scenarios.

apps/src/components/download/step-additional-details.tsx
renoirb: When debugging in production code, and in "development" (the way we currently do, for that matter) when we want to know what's the name of that component, it won't be a series of random letters and numbers

For example, the component `<S2DReleaseDate />` element in the following screenshot.

We can't really see in the HTML markup which piece of HTML is what. It makes it hard to debug, but we can make do by adding temporary class names.

<figure>

<figcaption>Screenshot of browser app with ReleaseDate component, a diff adding <tt>s2d-release-date-test</tt></figcaption>
</figure>

But in the code, the code we see in production, the name `S2DReleaseDate` the name in the code is something randomly generated, here in this example, it's `xT` that I've surrounded.

Reason-React-DisplayName-Debugging-1

apps/src/context/download-provider.tsx
renoirb: Good catch!

There's more to do, I'll push changes soon.

apps/src/context/download-provider.tsx
renoirb: True, the concept of `skipIndexes` with meaningless numbers is confusing. Particularly because the numbers in `DOWNLOAD_STEPS` are 1-indexed, and the `skipIndexes` are 0-indexed. Plus we lose completely the meaning. I should have had questioned more the model that implemented that changeset.

apps/src/lib/download/build-reset-payload-for-steps-after.ts
renoirb: What I planned doing is to keep that there for work context for the LLM agent. Then drastically reduce the comments at the time.

apps/src/context/download-provider.tsx
```diff
@@ -4,175 +4,160 @@
...
 	// Initialize URL sync
 	useDownloadUrlSync();
 
 	const [steps, setSteps] = useState<typeof STEPS>([...STEPS]);
 	// Start at step 2 if URL has variable parameter
 	const params = new URLSearchParams(window.location.search);
 	const hasVariable = params.has('var');
-	const [currentStep, setCurrentStepLocal] = useState<number>(hasVariable ? 2 : 1);
+	const [currentStep, setCurrentStepLocal] = useState<number>(
+		hasVariable ? DOWNLOAD_STEPS.variable : DOWNLOAD_STEPS.dataset,
```
renoirb, line 79: Here, that's much clearer what `hasVariable` or `(new URLSearchParams(window.location.search)).has('var')` changes in the steps.

apps/src/context/download-provider.tsx
renoirb: [Updated here](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/696/changes#r3539185124)

apps/src/context/download-provider.tsx
```diff
@@ -4,175 +4,160 @@
...
-				return STEPS.filter((_, index) => !skipIndexes.includes(index)) as unknown as typeof STEPS;
-			}
-			return [...STEPS];
-		});
+		setSteps(
+			() =>
+				STEPS.filter((_, index) =>
+					determineStepApplicable(climateVariable, index + 1)
+				) as unknown as typeof STEPS
+		);
```
renoirb, line 94: [replying to](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/696#discussion_r3514422418).

Much simpler like this.

apps/src/context/download-provider.tsx
renoirb: Yeah, it's [much simpler like this (link to how it's refactored)](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/696/changes#r3539207484), and to leverage the tests this PR already brings.

apps/src/lib/download/build-reset-payload-for-steps-after.ts
renoirb: yeah, `Path A` (or Option A) is confusing, I've refactored. And we'll clean up drastically comments at squash+merge

apps/src/lib/download/resolve-full-step-ordinal.test.ts
```diff
@@ -0,0 +1,127 @@
...
+	});
+
+	describe('a step skipped BEFORE the target shifts positions', () => {
+		// The AHCCD case: a station variable (not station_data) skips Variable
+		// Options and Additional Details, leaving:
+		//   [dataset, variable, location, sendRequest, result]
+		// Location is filtered position 3 but full ordinal 4. Passing the
+		// filtered position 3 straight into the full-space reset gate was the
+		// regression (4 > 3 wiped the Location being edited); the full ordinal 4
+		// restores the boundary (4 > 4 preserves it).
```
ChaamC: COmments shouldn't mention a regression introcued in some commits of this branch...

apps/src/lib/download/resolve-full-step-ordinal.test.ts
```diff
@@ -0,0 +1,127 @@
...
+		test('Send Request at filtered position 5 resolves to full ordinal 6', () => {
+			expect(resolveFullStepOrdinal(FULL_STEPS, filtered, 5)).toBe(6);
+		});
+
+		test('positions before the skip are unchanged', () => {
+			expect(resolveFullStepOrdinal(FULL_STEPS, filtered, 4)).toBe(4);
+		});
+	});
+
+	describe('boundary — the target step maps to itself, never past it', () => {
```
ChaamC: Do we really need that test? Aren't we already checking this at `filtered position %i resolves to full ordinal %i` one?

apps/src/lib/download/resolve-full-step-ordinal.test.ts
```diff
@@ -0,0 +1,127 @@
...
+			'dataset',
+			'variable',
+			'location',
+			'sendRequest',
+			'result',
+		] as const;
+
+		test('Location at filtered position 3 resolves to full ordinal 4', () => {
+			expect(resolveFullStepOrdinal(FULL_STEPS, filtered, 3)).toBe(4);
+		});
```
ChaamC: What is the point of this test considering it is already covered in the following test?

apps/src/lib/download/resolve-full-step-ordinal.test.ts
```diff
@@ -0,0 +1,127 @@
...
+			'filtered position %i resolves to full ordinal %i',
+			(filteredPosition, expectedOrdinal) => {
+				expect(
+					resolveFullStepOrdinal(FULL_STEPS, filtered, filteredPosition)
+				).toBe(expectedOrdinal);
+			}
+		);
+	});
+
+	describe('a step skipped FURTHER ALONG shifts only later positions', () => {
```
ChaamC: Not sure about this test case either, if it brings something more than what is tested at `filtered position %i resolves to full ordinal %i`.

I feel like the tests here should be more independent from our datasets and their own list of steps.
The tests should just verify the function.

Here it brings some dependency between the tests and our actual data. And what if our data and steps definition change, then we have some comments/tests which might become outdated here.

apps/src/lib/download/determine-applicable-steps.test.ts
```diff
@@ -0,0 +1,115 @@
...
+				1, 2, 4, 5, 6, 7,
+			]);
+		});
+
+		test('station "ahccd" — skips steps 3 and 5', () => {
+			const climateVariable = createStationVariable('ahccd');
+			expect(applicableOrdinals(climateVariable)).toEqual([
+				1, 2, 4, 6, 7,
+			]);
+		});
```
ChaamC: Isn't this test duplicated in the `describe('both station class names trigger the station skip rules', () => {` section below?
I would remove this one to keep the one below only.

apps/src/lib/download/build-reset-payload-for-steps-after.ts
ChaamC: I consider that anything that is put in a code review (unless in a draft state) should be in the state intended to be merged on the main branch. I don't think we should have comments kept for the context of the LLM agent in the review, as the reviewer cannot know what is intended for merging and it makes it harder to review.

ChaamC (changes requested):
I left some new comments, but great fix as the bug mentioned seems to be taken cared of now.

apps/src/lib/download/build-reset-payload-for-steps-after.ts
renoirb: Indeed.

I'll fix the redundant stuff.

apps/src/context/download-provider.tsx
renoirb: I found a way to make it clear what is a "Station Variable", it'll help clarify.

ChaamC (approved):
Let's merge this.
Tested this and didn't find any problem.

PR #698, opened 2026-06-16
CLIM-1413: fix missing delta value for return periods

ChaamC:
## Description

Delta values were missing from the grid cell popup. This is because charts data for return periods have a 15 year offset, and the delta value was selected with specific datetimes, in this case, the first year of every decade, like for other variables. This wasn't taking in account the offset of return periods values, and detected no data available for the requested year.

## Related ticket

[CLIM-1413](https://ccdpwiki.atlassian.net/browse/CLIM-1413)

[CLIM-1413]: https://ccdpwiki.atlassian.net/browse/CLIM-1413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/src/components/map-layers/raster-precalculated-climate-variable-values.tsx
+++ b/apps/src/components/map-layers/raster-precalculated-climate-variable-values.tsx
@@ -58,6 +58,7 @@ const RasterPrecalcultatedClimateVariableValues: React.FC<RasterPrecalcultatedCl
 		const variable = climateVariable?.getThreshold() ?? '';
 		const { lat, lng } = latlng;
 		const decadeValue = parseInt(dateRange[0]) - (parseInt(dateRange[0]) % 10) + 1;
+		const isReturnPeriod = climateVariable?.getClass() === 'ReturnPeriodClimateVariable';
 
 		const fetchData = async () => {
 			if (!decadeValue && !variableId) return;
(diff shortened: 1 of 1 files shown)


apps/src/components/map-layers/raster-precalculated-climate-variable-values.tsx
```diff
@@ -154,7 +154,9 @@ const RasterPrecalcultatedClimateVariableValues: React.FC<RasterPrecalcultatedCl
 				else if(frequency == 'winter') month = 11;
 			}
 
-			const timestamp = Date.UTC(decadeValue, month, 1, 0, 0, 0);
+			const isReturnPeriod = climateVariable?.getClass() === 'ReturnPeriodClimateVariable';
```
xfrenette-crim: This declaration (`const isReturnPeriod`) could be moved to line 61.

I know that the rest of this file doesn't follow this convention, but variables that are susceptible to be used for different logics (like, in this file, the variable id, the decadeValue, etc.) should be grouped at the top of the function. Reasoning: when developing new feature, we will generally first look in this group to see what is already available and what needs to be created.

xfrenette-crim (approved):
Approved with minor comment

PR #700, opened 2026-06-25
Minor return periods chart fix

ChaamC:
## Description

Minor fixes following partners feedback on return period data chart display:
- Trims grey plotband displayed on charts to be 10 years wide instead of 30 years
- Removes extra padding added at the beginning and end of x-axis
- Adapted to code to removed offset from the API, since it isn't needed anymore for the new chart format

### DIFF
--- a/apps/src/components/climate-data-chart.tsx
+++ b/apps/src/components/climate-data-chart.tsx
@@ -244,7 +244,7 @@ function getSeriesXValues(allSeries: Series[]): number[] {
 	return xValues;
 }
 
-const FIFTEEN_YEARS = 15 * 365 * 24 * 60 * 60 * 1000;
+const RETURN_PERIOD_PLOTBAND_YEAR_LENGTH = 9;
 
 /**
  * Component to render a chart using Highcharts with climate data.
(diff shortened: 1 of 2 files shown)


apps/src/components/climate-data-chart.tsx
xfrenette-crim, line 1: We can also remove the function `extendChartXAxis` that was created specifically to add the (now unnecessary) padding.

apps/src/components/climate-data-chart.tsx
```diff
@@ -383,24 +383,30 @@ const ClimateDataChart: React.FC<{
 		}
 
 		// Displayed period
-		let startYear = new Date(timestampKey).getUTCFullYear();
+		let tooltipStartYear = new Date(timestampKey).getUTCFullYear();
+		let plotBandStartYear = tooltipStartYear;
+		let tooltipYearLength = 29;
+		let plotBandYearLength = 29;
```
xfrenette-crim: `let` should be used only when a variable is explicitly redeclared, else it can stay a `const`. Here the variables `tooltipStartYear` and `tooltipYearLength` could stay `const`.


```suggestion
		const tooltipStartYear = new Date(timestampKey).getUTCFullYear();
		let plotBandStartYear = tooltipStartYear;
		const tooltipYearLength = 29;
		let plotBandYearLength = 29;
```

apps/src/components/climate-data-chart.tsx
```diff
@@ -544,9 +550,18 @@ const ClimateDataChart: React.FC<{
 
 		if(activeTab === TAB_VALUES.PERIOD_CHANGES) {
 			// Add plot band for reference period changes
+			let plotbandStartYear = 1971;
+			let plotbandEndYear = 2000;
+
+			if (isReturnPeriod) {
```
xfrenette-crim, line 518: `isReturnPeriod` must be added to the list of "dependencies" of the parent `useEffect`. It must be added in the list at line 569, after `activeTab`. Else, if `isReturnPeriod` changes, this code won't be re-executed.

xfrenette-crim (approved):

PR #702, opened 2026-06-25
CLIM-1409: Change Language Links

renoirb:
## What

Adds an EN⇄FR language switcher to the Map and Download SPA headers. The current language renders as inert text; the other language is an actual link that lands on the equivalent page in the counterpart locale, preserving the URL's query state.

## Design decisions

- **Explicit bidirectional origin map.**  `resolveAlternateOrigin` looks up the counterpart host from an explicit table (dev/uat en↔fr token pairs; prod bare-apex `climatedata.ca`↔`donneesclimatiques.ca`). `www.` is intentionally absent: it 301-redirects to the bare apex server-side before the SPA boots, so `window.location.hostname` is always the bare apex at runtime (verified live). An unrecognized host falls back to same-origin, keeping a usable link on localhost/staging.
- **Render.**  Current locale = inert `<span>` (with `aria-current` + endonym `aria-label`); other locale = real `<a href>` with `lang`/`hrefLang`.
- **State-derived query.**  The switch href is built from Redux state via memoized selectors (`selectMapUrlSearch` / `selectDownloadUrlSearch`), so it leads the debounced `replaceState` instead of lagging `window.location.search`.

## Implementation notes for reviewers

- **Shared serializers** (`apps/src/lib/url-params/`) are the single source for the URL query, used by both the url-sync writer and the switch selector — so the two cannot drift.
- **`currentStep` is a serialization gate**, never a URL value: `if step > 1 → set dataset/var, else clear both`. It must stay identical between writer and selector.
- **Byte-identical writer/selector (Download).** Both build from a fresh `URLSearchParams` (aligned with the Map writer). `dataset` + `var` is the full Download URL vocabulary.

## Testing

- New serializer regression tests (`build-map-url-params.test.ts`, `build-download-url-params.test.ts`) with `*.examples.ts` shared fixtures, including selector-vs-contract parity assertions that guard against future call-site drift. 
- `resolveAlternateOrigin` / `resolveAlternatePath` / `buildSwitchUrl` covered in `build-switch-url.test.ts` (both origin directions, prod apex pair, path table, query carry).
- Full suite: **507 passed | 4 skipped**; typecheck clean; lint 0 errors.

## Scope / follow-ups

- **Out of scope:** restoring deep Download-wizard step state across the switch (query is preserved; deep steps reset — acceptable). Tracked in CLIM-1410 #704 .
- **Pending stakeholder confirm:** that the current language should be non-interactive (implemented as such).


apps/src/lib/language-switch/build-switch-url.test.ts
```diff
@@ -0,0 +1,111 @@
...
+	resolveAlternateOrigin,
+	resolveAlternatePath,
+} from '@/lib/language-switch';
+
+describe('resolveAlternateOrigin (bidirectional host → counterpart lookup)', () => {
+	test.each([
+		['dev-en.climatedata.ca', 'https:', 'https://dev-fr.climatedata.ca'],
+		['dev-fr.climatedata.ca', 'https:', 'https://dev-en.climatedata.ca'],
+		['uat-en.climatedata.ca', 'https:', 'https://uat-fr.climatedata.ca'],
+		['uat-fr.climatedata.ca', 'https:', 'https://uat-en.climatedata.ca'],
```
ChaamC: Ces noms n'existent pas.
C'est plutôt `https://uat.donneesclimatiques.ca/` ou `https://uat.climatedata.ca/`

apps/src/lib/language-switch/build-switch-url.test.ts
```diff
@@ -0,0 +1,111 @@
...
+		['map', 'en', '/maps/'],
+		['map', 'fr', '/cartes/'],
+		['download', 'en', '/download/'],
+		['download', 'fr', '/telechargement/'],
+	] as const)('section %s + locale %s → %s', (section, locale, expected) => {
+		expect(resolveAlternatePath(section, locale)).toBe(expected);
+	});
+});
+
+describe('buildSwitchUrl (origin + path + query carry)', () => {
```
ChaamC, line 101: Should we add a test where the hostname is in the same language as the targetLocale?
e.g. : 
```
hostname: 'dev-en.climatedata.ca',
...
targetLocale: 'en',
```

apps/src/lib/language-switch/resolve-alternate-origin.ts
```diff
@@ -0,0 +1,52 @@
...
+ * shared apex and only flip the `en`/`fr` token; production crosses to the
+ * other registrable domain ({@link PROD_APEX_HOST_EN} ↔ {@link PROD_APEX_HOST_FR}).
+ * Explicit over a token-swapping regex: dead-obvious and exhaustively testable,
+ * one place for the truth. ([[LLM-Context-ClimateData-Ticket-CLIM-1409]], decision D1b.)
+ */
+const HOST_COUNTERPART = new Map<string, string>([
+	['dev-en.climatedata.ca', 'dev-fr.climatedata.ca'],
+	['dev-fr.climatedata.ca', 'dev-en.climatedata.ca'],
+	['uat-en.climatedata.ca', 'uat-fr.climatedata.ca'],
+	['uat-fr.climatedata.ca', 'uat-en.climatedata.ca'],
```
ChaamC: See other comment about these not existing.

apps/src/lib/language-switch/resolve-alternate-origin.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Explicit bidirectional host → counterpart-host lookup.
+ *
+ * Every known host maps to its other-language equivalent. Dev and UAT keep a
+ * shared apex and only flip the `en`/`fr` token; production crosses to the
+ * other registrable domain ({@link PROD_APEX_HOST_EN} ↔ {@link PROD_APEX_HOST_FR}).
+ * Explicit over a token-swapping regex: dead-obvious and exhaustively testable,
+ * one place for the truth. ([[LLM-Context-ClimateData-Ticket-CLIM-1409]], decision D1b.)
+ */
+const HOST_COUNTERPART = new Map<string, string>([
```
ChaamC: We could add support for :

[qa.climatedata.ca](https://qa.climatedata.ca/) and [qa.donneesclimatiques.ca](https://qa.donneesclimatiques.ca/)

and

[preprod.climatedata.ca](https://preprod.climatedata.ca/) and [preprod.donneesclimatiques.ca](https://preprod.donneesclimatiques.ca/)

apps/src/lib/url-params/build-download-url-params.test.ts
```diff
@@ -0,0 +1,66 @@
...
+		const scenario = downloadScenarioPastStepWithVariable();
+		const params = new URLSearchParams();
+
+		buildDownloadUrlParams(params, scenario.input);
+
+		const managedKeys = new Set(['dataset', 'var']);
+		for (const key of params.keys()) {
+			expect(managedKeys.has(key)).toBe(true);
+		}
+		expect(params.has('utm_source')).toBe(false);
```
ChaamC: Not sure about this test. Is `scenario` supposed to have the `utm_source` initially, and the `buildDownloadUrlParams` doesn't add it to the `params`?
I am just not sure if `utm_source` is even added anywhere here for the test, but correct me if I misunderstand.

apps/src/components/header.tsx
```diff
@@ -13,12 +17,19 @@ export default function Header(): JSX.Element {
...
-				<LogOut />
-			</a>
+			<div className="gap-x-2 px-4 py-2 flex items-center divide-x">
+				<a href="/" className="flex gap-2">
+					<span className="underline text-sm">
+						{__('Go back to the main website')}
+					</span>
+					<LogOut />
+				</a>
+				<HeaderLanguageLinks className="pl-4" />
```
ChaamC: I was wondering if it was possible to make the vertical separator more similar to what we have on other pages?

Currently proposed on the PR :

Image

What we have on the main page :

Image

So maybe something like :

Image

ChaamC (changes requested):
Good work over all, seems to work well in manual tests I did too. Good job.

Some minor comments to address.

apps/src/lib/language-switch/build-switch-url.test.ts
renoirb: ah, pourtant j'avais tenté de trouver dans Confluence. J'ajuste.

apps/src/lib/language-switch/build-switch-url.test.ts
renoirb: good question. There's a map in `apps/src/lib/language-switch/resolve-alternate-origin.ts` giving the alternate. It's a tuple of 2. We could have a 3rd where we tell the target if that's useful.


```ts
const HOST_COUNTERPART = new Map<string, string>([
	['dev-en.climatedata.ca', 'dev-fr.climatedata.ca'],
	['dev-fr.climatedata.ca', 'dev-en.climatedata.ca'],
	['uat.' + PROD_APEX_HOST_EN, 'uat.' + PROD_APEX_HOST_FR],
	['uat.' + PROD_APEX_HOST_FR, 'uat.' + PROD_APEX_HOST_EN],
	[PROD_APEX_HOST_EN, PROD_APEX_HOST_FR],
	[PROD_APEX_HOST_FR, PROD_APEX_HOST_EN],
]);
```

apps/src/lib/url-params/build-download-url-params.test.ts
renoirb: ah. Slipup. But this does nothing. And we don't have requirement for this. So I'll remove this.

apps/src/components/header.tsx
renoirb: It was maybe too subtile, but that's ... a bit too much, no?

Now? 
image

Remember: What's outside of maps and download isn't the same CSS and HTML at all.

apps/src/lib/language-switch/build-switch-url.test.ts
```diff
@@ -5,35 +5,77 @@ import {
...
+		['uat.donneesclimatiques.ca', 'https:', 'en', 'https://uat.climatedata.ca'],
+		['uat.climatedata.ca', 'https:', 'fr', 'https://uat.donneesclimatiques.ca'],
+		['qa.donneesclimatiques.ca', 'https:', 'en', 'https://qa.climatedata.ca'],
+		['qa.climatedata.ca', 'https:', 'fr', 'https://qa.donneesclimatiques.ca'],
+		['preprod.donneesclimatiques.ca', 'https:', 'en', 'https://preprod.climatedata.ca'],
+		['preprod.climatedata.ca', 'https:', 'fr', 'https://preprod.donneesclimatiques.ca'],
+		['climatedata.ca', 'https:', 'fr', 'https://donneesclimatiques.ca'],
+		['donneesclimatiques.ca', 'https:', 'en', 'https://climatedata.ca'],
+	] as const)(
+		'%s + locale %s → %s (sibling host)',
```
ChaamC, line 34: Just wondering about this string, is it formatted correctly?
Isn't the `locale` the 3rd argument in the array? Won't this display something like `donneesclimatiques.ca + locale https: -> en (sibling host)`?

If corrections are needed, review other similar strings in this file, which use similar formatting.

ChaamC (changes requested):

apps/src/lib/language-switch/build-switch-url.test.ts
renoirb: fixed.

ChaamC (approved):
Works well, good job!

PR #707, opened 2026-07-10
CLIM-1381: minor corrections on website texts

ChaamC:
- Minor corrections to french translations on the main site
- Added utility command to `dev.sh` script to compile the translations for the main site pages


dev.sh
```diff
@@ -208,6 +209,26 @@ function task-runner-shell {
   _docker_compose exec -it task-runner bash
 }
 
+function compile-main-site-translation {
```
xfrenette-crim: I'm noticing that this new function is based on the next `compile-react-apps-translation` function.

The execution of the function is good, but a few comments (which also apply to the `compile-react-apps-translation` function):

* The `echo`s (before and after execution) are not necessary since the functions in this script are only used to quickly execute development commands (except the help message). They don't need to add more explanation when executed. It would probably be better that all documentation be in the `docs/translate-the-site.md` file.
* The `docs/translate-the-site.md` could be updated to refer to the functions here.
* In the "Help" message of this script, we could create a new "Translation" section (and not use the "Other" section)
* We could add functions for the "update" and "generate" steps (see documentation) of the translations. Those steps are required when we add new translations. IMPORTANT: we have those steps only for the "main site", not for "React apps" (see documentation). Note: we should keep them as separate functions, since they each modify files that don't always need to be modified.
* To better show that all those functions are related to translation, we could prefix them with `i18n` (like the `wp i18n` command). For example:
  * `i18n-update-main-site` (or `i18n-main-site-update`)
  * `i18n-generate-main-site`
  * `i18n-compile-main-site`
  * `i18n-compile-react-apps`

dev.sh
```diff
@@ -42,7 +42,8 @@ function _show_help {
   echo ""
   echo "  Other:"
   echo "    download-docker-assets <URL>   Download the required Docker assets from the given URL."
-  echo "    compile-react-apps-translation Run translation files compilation for the React maps and download apps in apps/ folder"
+  echo "    compile-main-site-translation Run translation files compilation for the main site pages."
+  echo "    compile-react-apps-translation Run translation files compilation for the React maps and download apps in apps/ folder."
```
xfrenette-crim: The descriptions of the functions should be vertically aligned in a same section (if possible, even between sections, but not if a function in a section has a really long name). If possible, keep at least  two spaces before the description.


```suggestion
  echo "    compile-main-site-translation   Run translation files compilation for the main site pages."
  echo "    compile-react-apps-translation  Run translation files compilation for the React maps and download apps in apps/ folder."
```

xfrenette-crim (changes requested):

dev.sh
```diff
@@ -40,9 +40,14 @@ function _show_help {
   echo "  Task runner:"
   echo "    task-runner-shell       Start a shell on the 'Task runner' container."
   echo ""
+  echo "  Translation:"
+  echo "    i18n-update-main-site    Update the .pot translation file for the main site pages."
+  echo "    i18n-generate-main-site  Generate the .po translation file for the main site pages."
+  echo "    i18n-compile-main-site   Run translation files compilation for the main site pages."
+  echo "    i18n-compile-react-apps  Run translation files compilation for the React maps and download apps in apps/ folder."
```
xfrenette-crim: Suggestion:

```suggestion
  echo "    i18n-compile-main-site   Compile the .mo translation file for the main site pages."
  echo "    i18n-compile-react-apps  Compile the .mo translation file for the React maps and download apps."
```

docs/translate-the-site.md
```diff
@@ -34,6 +34,8 @@ When you add or modify a string in the PHP files, you need to update the
 > Note that this step is not required if you didn't add a new, or removed, a
 > new string to translate (skip to "Compiling translations").
 
+Either run the `./dev.sh i18n-update-main-site` command, or run the manual commands below :
```
xfrenette-crim: The step 1 below is required before executing this command (so this step should be step 2). Also, we don't need to keep the "manual steps" since they do the same as what the script does.

So the instructions could probably be simplified to (removing the "manual steps"):

(Same comment for the other sections in this file)

````markdown
1. Start the development environment.
2. Once the `portal` container is running, run:
   ```shell
   ./dev.sh i18n-update-main-site
   ```
````

xfrenette-crim (approved):
Approved with 2 minor comments.

PR #709, opened 2026-07-31
CLIM-1454: Map Download Image of the Current Map with Enriched Information

renoirb:
## Description

The map's download-as-image feature currently gives a PNG image with less information than a hand-made screenshot: the location popup isn't there, the legend isn't visible, and info that are not already in the location popup are omitted (e.g. only in the LocationModal not much info about when we have a grid, how much distance it covers). Most information is inlined in the LocationModal but is ... easy to miss, so we want it to be added as pills only when we're exporting as an image. And since this is an image, elements we normally interact with such as "See details", or the "x" to close a modal, or the +/- to change the zoom level are useless noise.

Things we already have in LocationPopup:

(following are how the current code behaves)

<p>
<figure>
CLIM-1454_Snapshot_2026-08-06_LocationPopup-Web
<figcaption>
<p>
How we normally see the LocationPopup, noitce the "summary" (that's specific to S2D) where we could also have a "See more" button.
</p>
</figcaption>
</figure>
</p>
<p>
<figure> 
CLIM-1454_Snapshot_2026-08-06_LocationPopup-Exported
<figcaption>
<p>
In the other image, we should not have the "Summary", the "i" *tooltips* as we can't do anything to open them on an image (or printed)
</p>
</figcaption>
</figure>
</p>

**Output this PR gives**

<p>
<figure> 
CLIM-1454_Snapshot_2026-08-06_Exported-3
<figcaption>
<p>
How it gets exported. (notice the "summary" and "i" weren't removed at the time of making that screenshot and were added last minute while writing this PR description)
</p>
</figcaption>
</figure>
</p>

<p>
<figure> 
CLIM-1454_Snapshot_2026-08-06_Exported-2

<figcaption>
<p>
When we compare scenarios
</p>
</figcaption>
</figure>
</p>

This PR makes the exported image display more visibly the following elements:

1. **Location popup** — added in the export by now supporting passing to the screenshot server where the marker (`markerLatLon`) was clicked, and the contents of the LocationModal as it was at the time of when the person clicked it (we can't reproduce exact same contents remotely, so we have to grab the HTML, and it can be 1 and at most 2 items when comparing scenarios and we split in two maps)
2. **Title pill** — composed for the export: climate variable + version
   (e.g. "Hottest Day — CMIP6"); on screen this exists as a breadcrumb, and a subtle gray line in a small font, so we're adding that information another time in the screenshot (displaying it 2x, not 3 as the breadcrumb at the top of the map isn't part of the exported image)
3. **Legend** — made sure it is expanded in the export so the colour ramp is visible.
4. **Grid-resolution pill** — export-only (e.g. "Grid ≈ 10x6 km"); the app
   never shows it on screen. -- Which should be the proper resolution, and only when it is showing a grid on the map (see `GridTypes`) 

### How it works

The SPA prepares the map for capture and signals readiness through a
`data-raster` attribute on the root element. The `climatedata-api` screenshot
service (Selenium) drives the page, waits for that signal, captures, and
returns the PNG. `$.fn.prepare_raster` is registered early
(`install-prepare-raster-stub.ts`) so the service can call it before the SPA
finishes mounting.

### Deliberately out of scope

- The raster **debugger** used during development — removed from this branch;
  may return as its own PR. If desired.
- The WordPress-side **same-origin raster proxy** — its own follow-up PR. If desired.
- Unit tests for the image-rastering modules — trimmed to keep this changeset
  reviewable; they exist on a side branch and can follow.
- A url-sync debounce finding uncovered while debugging — tracked separately.

## Related ticket

- CLIM-1454
- CanadianClimateDataPortal/climatedata-api#43


apps/src/main-map.tsx
```diff
@@ -1,3 +1,7 @@
+// Looking for `window.$.fn.prepare_raster`?
+//
+// Anything outside `apps/` relating to `prepare_raster` and Map Image Download
+// that uses the same name is effectively not used.
```
ChaamC, line 4: I think this comment shouldn't be here. No need to mention a specific function like `prepare_raster` on this high-level map-related file.

fw-child/apps/app-map.php
```diff
@@ -58,6 +58,19 @@
...
+	// screenshot contract -- on this page that assignment is the only one, with
+	// nothing to collide with or override.
+	//
+	// Why it cannot load: this template calls no wp_head()/wp_footer(), and
+	// WordPress fires wp_enqueue_scripts from inside wp_head() -- so nothing the
+	// theme enqueues reaches this page, including jQuery and the 'map-app'
+	// handle that carries map.js. (That isolation is the point; see the docblock
+	// above. Anything normally hooked to wp_head must be called directly, as
+	// add_favicon() is on the next line.)
+
```
renoirb:
```suggestion
	// **Caution:** Regarding `$.fn.prepare_raster` what's served as `https://climatedata.ca/maps/' (`/download`, and URLs in French) 
	// is a React Single-Page App (SPA) Frontend application the source is maintained from `<project root>/apps/` and does not include any files outside of it.
	// The use of `window.$.fn.prepare_raster` originated in  `fw-child/resources/js/map.js` but we do not use it and is effectively unused code that we plan to delete.
```

apps/src/components/map-layers/location-modal.tsx
```diff
@@ -60,14 +69,16 @@ const LocationModal = React.forwardRef<HTMLDivElement, LocationModalProps>(
...
 				{...props}
 			>
 				<button
 					className={cn(
 						'absolute top-3 right-3 text-gray-500 hover:text-gray-700 focus:outline-none'
 					)}
+					data-raster="false"
 					onClick={onClose}
-					aria-label="Close Modal"
+					aria-label={__('Close Modal')}
```
ChaamC, line 81: Why is this label translated?

fw-child/languages/react-apps/fr_CA.po
```diff
@@ -298,6 +298,9 @@ msgstr "ltr"
 msgid "Close panel"
 msgstr "Fermer"
 
+msgid "Close Modal"
+msgstr "Fermer"
```
ChaamC, line 302: See comment about translation in `location-modal.tsx`

apps/src/lib/location-modal-class-names.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Shared Tailwind class-name lists for the LocationModal's outer element,
+ * consumed by both the real `LocationModal` component and
+ * `prepare-raster.ts`'s hand-built look-alike `<div>` (outside React) that
+ * replays the popup for the screenshot service.
+ *
+ * Hoisted here instead of duplicated as string literals: a duplicated
+ * literal can drift — e.g. `md:left-16` typo'd into two junk classes —
+ * silently repositioning the raster replay to the horizontal centre of the
+ * map instead of the intended top-left corner.
```
ChaamC, line 10: We could remove that comment, it might not be necessary to document why we want to avoid duplicating code.

apps/src/components/map-info/download-map-modal.tsx
```diff
@@ -46,51 +44,103 @@ const DownloadMapModal: React.FC<{
...
+				 	locationPopupHtml,
+					markerLatLon,
+				}
+			}
+			// No `?? ...` fallback needed: `resolveSignalReady` ships as an identity
+			// function, returning the `signalRasterReady` passed to it.
+			const signalReady = resolveSignalReady(signalRasterReady);
+
+			// Do not signal readiness after a preparation failure. We let its wait time out.
+			// It means the `.to-raster` className to trigger screenshot and it should have had waited longer.
```
ChaamC: Unclear formulation?
I'm just not sure what it means, and I think the sentence can be reformulated differently.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+export const isGridType = (
+	value: unknown,
+): value is GridType =>
+	typeof value === 'string' && GRID_TYPE_VALUES.has(value);
+
+export const isClimateVariable = (
+	climateVariable: ClimateVariableInterface | null,
+): climateVariable is ClimateVariableInterface =>
+	typeof climateVariable === 'object' &&
+	// Just doing what's here is probably too permissive.
```
ChaamC, line 268: Unsure about this comment if it should be left here.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * Whether a value is a grid identity this module carries a resolution label for.
+ */
+export const isGridType = (
+	value: unknown,
+): value is GridType =>
+	typeof value === 'string' && GRID_TYPE_VALUES.has(value);
+
+export const isClimateVariable = (
```
ChaamC, line 264: I am unsure about this constant if it is truly needed or not, and if the name is representative or not.

It seems to just be used in `isStationClimateVariable` to check if the variable has a `getId()` function that returns a string.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+	climateVariable: ClimateVariableInterface | null,
+): climateVariable is ClimateVariableInterface =>
+	typeof climateVariable === 'object' &&
+	// Just doing what's here is probably too permissive.
+	typeof climateVariable?.getId() === 'string';
+
+export const isStationClimateVariable = (
```
ChaamC, line 271: Is this the best file for this function, since it doesn't relate explicitly to grids? Same thing for the `isClimateVariable` constant.

But maybe okay here, since it is only used here.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+ * show.
+ *
+ * @remarks
+ * Two unrelated situations yield `null`, and neither is an error:
+ * - the variable is station data, which is measured at points and has no grid;
+ * - the variable declares no grid type, or one this module carries no label for.
+ *
+ * Display is otherwise not conditioned on how the map is being browsed. The sizes are
+ * declared per data source, so a variable shown by region still has the grid its data was
+ * produced on, and still states it.
```
ChaamC, line 301: I didn't test it, but I wonder about screenshots that don't use grids (Census subdivisions, health regions, watersheds). They don't use grids. So, it would make sense not to display the grid info on the screenshots for those cases.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+ */
+export const getGridTypeFor = (
+	climateVariable: ClimateVariableInterface | null,
+): GridType | null => {
+	if (isStationClimateVariable(climateVariable)) {
+		return null;
+	}
```
ChaamC, line 308: Doesn't the `station-climate-variable.ts` implemented an explicit `getGridType()` function that would already return `null`?

Maybe this code is not needed.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+ * S2D shares this label because `S2DClimateVariable.getGridType()`
+ * reports the same `canadagrid` lattice a forecast is selected on.
+ */
+const GRID_RESOLUTION_LABEL_STATISTICALLY_DOWNSCALED_AND_S2D =
+	'~10×6km' as const;
+
+/**
+ * 11×7km : Marine Projections
+ *
+ * Source: (to confirm)
```
ChaamC, line 111: Can we remove this comment, which doesn't really add valuable information?
Same thing for other identical comments below.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+		GRID_RESOLUTION_LABEL_SPEI,
+	[GridTypes.ERA5LANDGRID]:
+		GRID_RESOLUTION_LABEL_DAYS_WITH_HUMIDEX_ABOVE_THRESHOLD,
+	[GridTypes.SLRGRID]:
+		GRID_RESOLUTION_LABEL_MARINE_PROJECTIONS,
+	[GridTypes.SLRGRID_CMIP6]:
+		GRID_RESOLUTION_LABEL_MARINE_PROJECTIONS,
+	/**
+	 * See {@link GRID_RESOLUTION_VALUE_MARINE_PROJECTIONS} — anisotropic, unlike every
+	 * other entry here.
```
ChaamC, line 233: This info is already written in the `GRID_RESOLUTION_VALUE_MARINE_PROJECTIONS` description, no need to repeat it.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+ *   Published as '0,1° (environ 11 km de latitude, 2-8 km de longitude)'
+ *   ({@link https://donneesclimatiques.ca/le-portail/nos-donnees/#sea_level | Changement relatif du niveau de la mer})
+ *   — French only; the English page states no figure for this dataset.
+ *
+ * - `?var=allowance` : Vertical Allowance.
+ *   Published as '0.1° (approximately 11 km latitude, 4-8 km longitude)'
+ *   ({@link https://climatedata.ca/about/our-data/#vert_allowance | Vertical Allowance})
+ *   — anisotropic, unlike every other entry in this table. WFS-measured.
+ *
+ * 0.1° for all three, coastal-only.
```
ChaamC, line 174: Unneeded comment

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -0,0 +1,313 @@
...
+ * `climateVariableId`s:
+ * - `?var=sea_level` : Relative Sea-Level Changes.
+ *   Published as '0,1° (environ 11 km de latitude, 2-8 km de longitude)'
+ *   ({@link https://donneesclimatiques.ca/le-portail/nos-donnees/#sea_level | Changement relatif du niveau de la mer})
+ *   — French only; the English page states no figure for this dataset.
+ *
+ * - `?var=allowance` : Vertical Allowance.
+ *   Published as '0.1° (approximately 11 km latitude, 4-8 km longitude)'
+ *   ({@link https://climatedata.ca/about/our-data/#vert_allowance | Vertical Allowance})
+ *   — anisotropic, unlike every other entry in this table. WFS-measured.
```
ChaamC, line 172: Not really sure what it means and where it is coming from.

apps/src/components/map-info/README.md
```diff
@@ -0,0 +1,9 @@
+# `components/map-info/`
+
+Panels and modals describing the map the user is currently looking at.
+
+Not a complete inventory of the folder; entries are added as files are documented.
+
+- [`download-map-modal.tsx`](./download-map-modal.tsx) → `DownloadMapModal` — the modal behind the Download button.
+  - `handleDownloadClick` — POSTs the current view to the screenshot service, then saves the returned PNG as the downloaded map image.
+  - Registers `window.$.fn.prepare_raster`, the entry point the screenshot service's headless browser calls. See [`lib/map/image-rastering/`](../../lib/map/image-rastering/) for the raster vocabulary.
```
ChaamC, line 9: Sure, we can add READMEs to give some explanations about a folder's content, but I don't think we should document specific functions contained in each file.  Function and other file-specific details should be documented in the file itself, if needed.

apps/src/lib/map/image-rastering/types.ts
```diff
@@ -0,0 +1,71 @@
...
+ * The function we attach to `window.$.fn.prepare_raster` that is called by the server-side
+ * screenshot service (`climatedata-api`, `climatedata_api/raster.py`) in a headless browser
+ * to prepare the map page for a screenshot.
+ */
+export type Prepare_Raster = (
+	locationPopupHtml?: PrepareRasterPostHttpPayload['locationPopupHtml'],
+	markerLatLon?: PrepareRasterPostHttpPayload['markerLatLon'],
+) => void;
+
+export type PrepareRaster = (
```
ChaamC, line 67: Could be pertinent to rename this type, and/or the above one. `Prepare_Raster` and `PrepareRaster` types are 2 distinct types, but the names are almost the same. Maybe adding a relevant suffix to distinguish their individual role?

apps/src/lib/map/image-rastering/resolve-signal-ready.ts
```diff
@@ -0,0 +1,11 @@
+import type { SignalReady } from './signal-raster-ready';
+
+/**
+ * Returns the readiness callback `prepareRaster` should use — the seam a
+ * caller substitutes its own through. Ships as an identity function.
+ */
+export const resolveSignalReady = (
```
ChaamC: What is the point of this function?
Is there an architectural concern or a need to expand on this in the future?
Or, if not, is this overengineered and could be simplified?

apps/src/lib/map/image-rastering/wait-for-maps-settled.ts
```diff
@@ -0,0 +1,127 @@
...
+ * before the map counts as settled.
+ *
+ * More than one, because the two panes are synchronised in both directions (see
+ * the `sync` calls in `map.tsx`): a move on one pane propagates to the other, so
+ * the second pane starts requesting its tiles slightly *after* the first pane
+ * has finished with its own. A single idle reading can land in that gap and
+ * report a map that has not finished. Requiring consecutive idle frames narrows
+ * the window. It does not close it, and no finite number would.
+ *
+ * This value and the original 8000ms single-call timeout were introduced together
```
ChaamC, line 39: What is the original 8000ms timeout mentioned here?

apps/src/lib/map/image-rastering/prepare-raster.ts
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * Ends by calling `signalReady`, the injected readiness signal — defaults to
+ * {@link signalRasterReady} — which is how the service learns it may proceed.
+ * Everything awaited before that point is something the screenshot depends on
+ * *and* that the platform can report the completion of. Nothing here waits out
+ * a fixed delay.
+ */
+// Exposed globally as $.fn.prepare_raster because a server-side headless-browser
+// screenshot service invokes that exact expression against the page. A second,
+// unrelated function under the same name lives in fw-child/resources/js/map.js;
```
ChaamC, line 36: Just a comment, I think that other function is just some code remaining from the v1, that is now unused.

ChaamC (commented):
It would be interesting to have the Legend on the right map when using the comparison mode.

Also, since I haven't been able to test it, considering of the complexity of testing with a local instance, I am writing here a list of some use cases to cover and check :
- Statistically Downscaled Global Climate Projections variable
  - Days with Humidex above threshold variable (uses adifferent grid)
  - SPEI variables (uses a different grid)
- S2D variables
- Marine Climate Projections variables
- Climate Normals (should have less elements in the screenshot, for example, no grid. The variable name should still be displayed at the top of the screenshot.)
- Comparing scenarios
- Mobile format (which I guess shouldn't change the expected format since we are using a second headless browser versin of the site on the API server)
- Validate that changing a parameter (like interactive regions, frequencies, scenarios, delta values, map colours, opacity, etc.) is kept on the screenshot too
  - I think this was already working before this task, so it's just to check that it didn't break anything.

apps/src/main-map.tsx
renoirb: OK. But not having this explanation had the agent when I asked to analyze go to a tangent that had me lose hours. But that's not the point of this PR.

apps/src/components/map-layers/location-modal.tsx
renoirb: Because the label is useful to tell (to assistive tech who can't "*see*" -- screen-readers) how to dismiss something in focus (and hiding the rest) and how to move it out.

Quoting [<span title='Mozilla Developer Network'>MDN</span>'s \[aria-label\] page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label)

> Sometimes, the default [accessible name](https://developer.mozilla.org/en-US/docs/Glossary/Accessible_name) of an element is missing or the accessible name does not accurately describe the contents of the element and there is no content visible in the DOM that can be associated with the object to give it meaning. A common example of such an element is a button containing an SVG icon without any text.

That label wasn't translated, so in french it would "read" (hear) with a french pronunciation english word.

But, maybe it could simply be removed. I didn't test with most popular screen reader this feature.

Regarding aria, it's like a vorpal sword. It cuts, but it can hurt. [Here's an humorous video explaining what it's for (and entertaining)](https://briefs.video/videos/what-is-aria-even-for/); sometimes it can get so complicated that we want *aria* to *die*... "*die* *aria*".

apps/src/components/map-info/download-map-modal.tsx
renoirb: The className "to-raster" is the way to communicate it is ready.

renoirb (commented):
I had these comments unsubmitted. Submitting them so they can be found.

fw-child/apps/app-map.php
renoirb: no longer relevant.

apps/src/lib/map/image-rastering/resolve-signal-ready.ts
renoirb: You are right, it was over-engineered. It is gone now.
 
`resolveSignalReady` was a leftover from the raster debugger. I had added a `signalReady` parameter to prepareRaster so that debugger can do things there.
The debugger was removed earlier on this branch. 
`resolveSignalReady` stayed behind as if it were required. It is not, and as you spotted, it does not earn its place in the codebase. So I removed it.

docs/developing-with-the-screenshot-service.md
renoirb: Testing procedure.

apps/src/lib/map/image-rastering/create-fetch-target-to-raster-with-encoded-url.ts
```diff
@@ -16,6 +16,6 @@ export const createFetchTargetToRasterWithEncodedUrl = (
 		mapUrl.toString(),
 		window.URL_ENCODER_SALT,
 	).encoded;
-	const rasterEndpoint = new URL('/raster', window.DATA_URL);
+	const rasterEndpoint = new URL('/raster', 'https://dev-en.climatedata.ca:5001');
```
ChaamC: This change shouldn'd be committed on the repo.

docs/developing-with-the-screenshot-service.md
```diff
@@ -0,0 +1,326 @@
...
+an error.
+A `502` means that the proxy cannot reach the service. See
+[The proxy cannot reach the screenshot service](#the-proxy-cannot-reach-the-screenshot-service).
+Any other error comes from the service itself, so read its output in the
+terminal where it runs.
+
+Then open the downloaded image. It must carry the four elements: the
+title, the legend, the grid, and the location popup with its marker. It can
+also show a scenario pill, which was already in exported images before this
+work.
```
ChaamC: I think this documentation should be more generic, no need to reference this specific feature we are adding.

docs/developing-with-the-screenshot-service.md
```diff
@@ -0,0 +1,326 @@
...
+
+The browser sends the inner HTML of the popup to the service. The close button
+and the "See details" link of the popup carry `data-raster="false"`, so neither
+appears in the image.
```
ChaamC: Should this level of detail be here, or should be contained in the actual code? This seems closer to implementation details than to testing instructions for the screenshot service.

docs/developing-with-the-screenshot-service.md
```diff
@@ -0,0 +1,326 @@
...
+the hardcoded path. No `sudo` and no system package are needed.
+
+The `selenium-manager` binary ships inside the installed `selenium` package. It
+downloads a chromedriver that matches your installed Chrome. Its path contains
+the Python version and the platform, so find it with `find`.
+
+1. From the root of the `climatedata-api` clone, ask `selenium-manager` for a
+   driver that matches the installed Chrome:
+   ```shell
+   SELENIUM_MANAGER=$(find .venv -name selenium-manager -type f)
```
ChaamC: For this line, I had to use `SELENIUM_MANAGER=$(find .venv -path '*/linux/selenium-manager' -type f)`, because the suggested command returned 2 different paths for me, one with `*/linux/selenium-manager` and another with `*/macos/selenium-manager`.

Not sure, how it is configured on your environment, may I suggest we use `SELENIUM_MANAGER=$(find .venv -path '*/linux/selenium-manager' -type f)` for a more precise output?

docs/developing-with-the-screenshot-service.md
```diff
@@ -0,0 +1,326 @@
...
+       extra_hosts:
+         - "host.docker.internal:host-gateway"
+       volumes:
+         - type: bind
+           source: dockerfiles/mounts/config/raster-proxy.conf
+           target: /etc/nginx/conf.d/raster-proxy.conf
+       ports:
+         - "5001:5001"
+   ```
+   If you do not use Ladle, leave out the `task-runner` entry.
```
ChaamC: I think I would leave out this remark about Ladle.

I would also remove this part in the example :
```
     task-runner:
       ports:
         - "61000:61000"  # Ladle development server
```

This configuration part is not related to testing the screenshot service.
We could potentially just put instead :
```
     ...
```
to represent any other potential configs already used by the developer in his override file.

ChaamC (changes requested):
Left some minor comments on the documentation file, but it was working well to test the screenshots feature locally.

I also tested the different test cases I had listed ([here](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#pullrequestreview-4846158139)), here are some remarks, to address :
- Marine Climate Projections : info banner is displayed. I think it would be relevant to hide it in the screenshot, especially since it shows a "Read more" link. Probably easy to fix, by just adding the html raster tag with the right value.

- Scenario comparisons : Legend displayed only on Left (not only in screenshots, but also on the Maps page). It would be better to display it on the map positioned to the right. Only do this change if it is an easy one... A thing to keep in mind is that it affects the display of the map page, not just on the screenshot. So, I would rather not degrade the display with our update.

- Error 400 on some requests :
  - on the Climate Normals dataset
  - if no location popup is opened (on any dataset)
Tell me if I am wrong, but I think those 400 errors are the API's responsibility. I think it's for cases when only the `url` is sent (no `locationPopupHtml` or `markerLatLon`). I will have a check on the API, so that it still supports this case.

renoirb:
[Replying to](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#pullrequestreview-5223204453)

(...)
> 
> I also tested the different test cases I had listed ([here](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#pullrequestreview-4846158139)), 

(I'll reply to those from that message)


> (...) here are some remarks, to address :
> 
> - Marine Climate Projections : info banner is displayed. I think it would be relevant to hide it in the screenshot, especially since it shows a "Read more" link. Probably easy to fix, by just adding the html raster tag with the right value.

- [ ] Not done yet. Should be trivial.

> - Scenario comparisons : Legend displayed only on Left (not only in screenshots, but also on the Maps page). It would be better to display it on the map positioned to the right. Only do this change if it is an easy one... 

This is solved. It was simpler than I thought.

>   (Scenario comparisons...) A thing to keep in mind is that it affects the display of the map page, not just on the screenshot. So, I would rather not degrade the display with our update.

&lt;aside&gt;(**Web Historian nerd hat:** That's what people called back around 2012 "*responsive design*" and later on *media-queries* to tell how to adjust stuff based on available space based on screen size.)&lt;/aside&gt;

The earlier job (before I've joined and we've taken charge of the site back ~ +>12 months ago.) the mobile screen size when asking to download image on mobil (BRB, going to check in production...) 

> (...)
> - Error 400 on some requests 😀 

> 
>   - on the Climate Normals dataset
> 
>   - if no location popup is opened (on any dataset)
> 
> 
> 
> EDIT : These 2 error 400 cases should be fixed by my [latest commit](https://github.com/CanadianClimateDataPortal/climatedata-api/commit/7d1c87a559c578d01ffa98f195544c6c76bceb37) on the API, which now supports POST request that don't have a JSON payload, like in those cases.

apps/src/components/map-info/README.md
renoirb: You want me to remove the README.md that I've added? I worked to make sure the README are with least amount of noise.  In rendered form it currently gives for [apps/src/components/map-info/README.md](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/blob/42778a2d/apps/src/components/map-info/README.md) (link with a hash snapshot) to [apps/src/lib/map/image-rastering/](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/tree/42778a2d/apps/src/lib/map/image-rastering)

apps/src/lib/map/image-rastering/types.ts
renoirb: I've renamed the type `Window_Fn_Prepare_Raster` so it's not simply `PrepareRaster` and `Prepare_Raster`.

apps/src/lib/map/image-rastering/resolve-signal-ready.ts
renoirb: This is removed.

apps/src/components/map-info/README.md
ChaamC: No, my comment was only about line 8-9. I think they shouldn't be in this file and that it makes it too verbose to document individual functions in this README.

apps/src/lib/map/image-rastering/wait-for-maps-settled.ts
renoirb: It was from a past commit it won't be seen in an earlier implementation. Removed.

apps/src/lib/map/image-rastering/prepare-raster.ts
renoirb: Yeah, that's noise I should've had removed.

docs/developing-with-the-screenshot-service.md
renoirb: That's removed.

docs/developing-with-the-screenshot-service.md
renoirb: The issue is that we may get more than one.
But showing how it looks like would be helpful, I'll add that.

We have to be more generic than only Linux, too.

docs/developing-with-the-screenshot-service.md
renoirb: fixed.

apps/src/components/map-info/README.md
renoirb: Right.

But that's more for agents than us (except we're not giving in on a vendor in particular). We can read large pices of code and navigate with an IDE.

This file is definitely smaller than getting to the function name and all the surrounding code and the imports.

apps/src/components/map-layers/location-modal.tsx
renoirb: also. "Close Modal" in french, and screen reader (i.e. reading the text) would pick `aria-label=` and read with french pronunciation something in english. A bit like "*Les Gants Élégants*" (search on youtube)

docs/developing-with-the-screenshot-service.md
```diff
@@ -0,0 +1,335 @@
...
+```shell
+git checkout -- apps/src/lib/map/image-rastering/create-fetch-target-to-raster-with-encoded-url.ts
+```
+
+## Download a map image
+
+1. Open <https://dev-en.climatedata.ca/maps/>, and move the map to any area.
+2. Click a location on the map, so that its popup opens.
+4. In the map header, click the red 'download' button, it will open a modal
+5. Under the title **Download image from viewport**, click the
```
ChaamC:
```suggestion
3. In the map header, click the red 'download' button, it will open a modal
4. Under the title **Download image from viewport**, click the
```

docs/developing-with-the-screenshot-service.md
```diff
@@ -0,0 +1,335 @@
...
+reads the `SE_CHROMEDRIVER` environment variable, and that variable overrides
+the hardcoded path. No `sudo` and no system package are needed.
+
+The `selenium-manager` binary ships inside the installed `selenium` package. It
+downloads a chromedriver that matches your installed Chrome. Its path contains
+the Python version and the platform, so find it with `find`.
+
+1. From the root of the `climatedata-api` clone, ask `selenium-manager` for a
+   driver that matches the installed Chrome:
+   - On a development machine or VM a past run of Selenium with Chrome
```
ChaamC:
```suggestion
   - On a development machine or VM, a past run of Selenium with Chrome
```

apps/src/lib/map/image-rastering/create-fetch-target-to-raster-with-encoded-url.ts
ChaamC: bump

ChaamC (changes requested):

apps/src/lib/vocabulary/grid-resolution.ts
ChaamC: Bump
Tested it, and the grid element is displayed on those types of interactive regions, although there is no proper grid on those.

docs/developing-with-the-screenshot-service.md
renoirb: done commit 3be0ef3e

apps/src/lib/map/image-rastering/create-fetch-target-to-raster-with-encoded-url.ts
renoirb: done 93f8380c

docs/developing-with-the-screenshot-service.md
renoirb: Resolved 3be0ef3e

apps/src/lib/location-modal-class-names.ts
renoirb: Solved 5d01dfb9

apps/src/main-map.tsx
renoirb: Solved 33feefcd

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: yeah. TIL "*anisotropic*". I've removed it.  I like the mention of the link to the source, and how they relate in the map though.

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: ok, removed too.

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: Could you weigh a compromise here? Either we keep the simple guard and I remove the comment, or we bring back a stricter lookup. The context is below.

I left that comment on purpose, to see whether its reason would come up.

The reason I've added that comment is that `typeof climateVariable?.getId() === 'string'` is a bit pedantic, it's almost the same as doing `true === true`.

In this PR, we needed a type guard, so that `isStationClimateVariable` can narrow the type before `getGridTypeFor` uses it.

I did write a stricter version first (cd72499f), and moved away from it (abfeefe9):

```ts
export const isClimateVariable = (
	climateVariable: ClimateVariableInterface | null,
): climateVariable is ClimateVariableInterface =>
	typeof climateVariable === 'object' &&
	// Just doing what's here below only would be too permissive.
	typeof climateVariable?.getId() === 'string' &&
	// Maybe this is too much?
	ClimateVariables.find((c) => c.id === climateVariable.getId()) !== undefined;
```

I judged the lookup in `ClimateVariables` to be overkill for what `getGridTypeFor` needs. It also has a cost in placement: `config/climate-variables.config.ts` imports `GridTypes` from `lib/vocabulary`, so putting the lookup on either side creates an import cycle.

Which do you prefer?

1. Keep the simple guard as it is, and I remove the comment.
2. Bring back the stricter lookup, and we decide where it lives without the cycle.

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: See [related](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#discussion_r4076917953)

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: You are right, and I missed this detail.

Before this change (b898cc29), the grid size depended only on the variable's data source.
So with `&region=census` (or health regions, watersheds), the map drew region shapes, but the pill still showed `Grid= ~10×6km` whenever the variable's grid type was a gridded one.

Now the pill shows the grid size only when the interactive region is `gridded_data`, meaning when grid cells are actually on the map.
For every other region, it shows no grid size.

- `?var=s2d_precip_accum&dataset=260` -> `~10×6km`
- `?var=hottest_day&dataset=216&region=census` -> `''` (i.e. no longer showing the pill)

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: I trusted that the code with the values existed, and I clicked all of them, and they aren't visible on the site both in production or in dev, only `msc_climate_normals`.

You are right. Every station variable uses the string value `StationClimateVariable` or its "*subclass*", and getGridType() returns null there, it changes nothing.

I removed it, with `isStationClimateVariable`, `isClimateVariable` and the station id set that only served it. 

Fixed in f0ce21e0

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: Removed in f0ce21e0, together with the station check (see [r3737193728](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#discussion_r3737193728)).
This replaces the two options I offered above.

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: Removed in f0ce21e0, together with the station check (see [r3737193728](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#discussion_r3737193728)).

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: No longer applies: the guards are removed in f0ce21e0 (see [r3737193728](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/709#discussion_r3737193728)).

apps/src/components/map-info/download-map-modal.tsx
renoirb: See 4823ff66

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: See 54f45724e

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: See 54f45724e

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: See 54f45724e

apps/src/components/map-info/README.md
renoirb: You are right. Both lines are now gone in a4df6f00.

renoirb (commented):
That should do it.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -161,24 +152,20 @@ export const GRID_RESOLUTION_VALUE_STATISTICALLY_DOWNSCALED_AND_S2D =
  * 11×7km : Marine Projections
  *
  * `climateVariableId`s:
- * - `?var=sea_level` : Relative Sea-Level Changes.
+ * - `?var=sea_level` : Relative Sea-Level Changes (CMIP5).
```
ChaamC: No need to add the `CMIP5` mention. There is data for both CMIP5 and CMIP6 for this variable.

apps/src/lib/vocabulary/grid-resolution.ts
```diff
@@ -161,24 +152,20 @@ export const GRID_RESOLUTION_VALUE_STATISTICALLY_DOWNSCALED_AND_S2D =
...
  *   Published as '0,1° (environ 11 km de latitude, 2-8 km de longitude)'
  *   ({@link https://donneesclimatiques.ca/le-portail/nos-donnees/#sea_level | Changement relatif du niveau de la mer})
- *   — French only; the English page states no figure for this dataset.
  *
  * - `?var=allowance` : Vertical Allowance.
  *   Published as '0.1° (approximately 11 km latitude, 4-8 km longitude)'
```
ChaamC: I am not sure we really need to give those latitude/longitude details about kms.

I don't mind if it stays though, but if it does, please correct the longitude. It should be the same as the `sea_level` var above : `2-8 km`.

ChaamC (commented):
All good for the remaining comments!

I just added 2 very minor corrections on some comments.

apps/src/lib/vocabulary/grid-resolution.ts
renoirb: Ah. That info.  That's because when I did ask claude to do the research for me, I asked to check on multiple URLs and use contents from both languages. And it noted that there's a discrepancy in the same information but in a different language. And it is reminding me constantly with this example.

We could tell our client to adjust this discrepancy

Screenshot from 2026-09-23 15-13-51

ChaamC (approved):

apps/src/lib/vocabulary/grid-resolution.ts
ChaamC: I just corrected the info on the website. Thanks for pointing it out!

PR #711, opened 2026-08-17
CLIM-1447: S2D-Maps Add Decadal Time Periods Display Support

renoirb:
## Description

- Add the new frequencies to the **Frequency** drop-down menu:
  CLIM-1447_Original-Ticket-Illustration-1
    - Decadal (5 years, Annual)
    - Decadal (5 years, May-Sep)
    - Decadal (5 years, Nov-Mar)
     
- Update the time slider to show Decadal periods, such as `2026-2030` and `2031-2035`:
  CLIM-1447_Original-Ticket-Illustration-2
    - Hide or disable the time slider when **Climatology** and a Decadal frequency are selected. Climatology uses the same data for both time periods.
    - The time-slider years must follow the Decadal data release date. For a 2026 release, the values must be `2026-2030` and `2031-2035`.
    - The slider has only two values. It must move fully left or fully right.
    - For a 2026 release:
        - Values below the slider bar: `2026` and `2035`
        - Values above the slider bar: `2026-2030` and `2031-2035`
     
- Revise the pop-up panel content for the new Decadal data:
  CLIM-1447_Original-Ticket-Illustration-3
    - The screenshot comes from the partner and predates the latest pop-up updates. Summary and the information pop-up remain relevant and must support Decadal data.

The new Decadal frequencies use these keywords for the API and GeoServer:
- Decadal (5 years, Annual): `decadal-ann`
- Decadal (5 years, May-Sep): `decadal-may-sep`
- Decadal (5 years, Nov-Mar): `decadal-nov-mar`

## Related ticket

- CLIM-1447


renoirb:
CLIM-1447_Snapshot_2026-08-18-19-03-00
CLIM-1447_Snapshot_2026-08-18-19-03-01
CLIM-1447_Snapshot_2026-08-18-19-04-14

apps/src/components/s2d/frequency.tsx
```diff
@@ -0,0 +1,96 @@
...
+		...restProps,
+	};
+
+	// BEGIN: We probably DO NOT WANT THIS (no we don't!)
+	for (const option of FrequencyField.options) {
+		if (/Annuelle/.test(option.label)) {
+			option.label = option.label.replace('Annuelle', 'annuelle');
+		}
+	}
+	// END: We probably DO NOT WANT THIS
```
renoirb: We can't change 'Annuelle' to 'annuelle' in translation because we already have 'Annual' into 'Annuelle' elsewhere. And the other solution is... not a good one that looks like this.

Image

renoirb:
OK @ChaamC this should be ready now.  And the 2nd PR is #712

apps/src/lib/climate-variable-base.ts
```diff
@@ -524,6 +524,7 @@ class ClimateVariableBase implements ClimateVariableInterface {
 	}
 
 	getLayerValue(scenario: string | null | undefined, section?: string): string {
+		// @TODO Check effect for all Decadal variants
```
ChaamC: This comment should probably not be kept.

fw-child/languages/react-apps/fr_CA.po
```diff
@@ -1190,8 +1195,17 @@ msgstr "Les périodes disponibles changeront si une fréquence différente est s
...
+msgstr "Décennale (%s)"
+
+msgid "Decadal %s probability:"
+msgstr "Probabilité décennale pour \"%s\""
+
+msgid  "May-Sep"
+msgstr "mai-sep"
+
+msgid  "Nov-Mar"
+msgstr "nov-mar"
```
ChaamC: If we cannot have the "annuelle" display value with the lower case (or if it requires too much work or shady code), I would choose to harmonize those and use the upper case for those translations too, to avoid mixed casing usage (like in picture below). We can see if that works with them afterwards. No need to put too much time on this, since this might be a minor detail.

Image

apps/src/lib/s2d.ts
```diff
@@ -195,9 +269,10 @@ export function buildSkillLayerName(
 	releaseDate: Date
 ): string | null {
 
+	// @TODO Fix should probably NOT `null` for every Decadal frequency.
```
ChaamC: Can this be fixed or removed?

apps/src/lib/s2d-climate-variable.ts
```diff
@@ -116,22 +117,30 @@ class S2DClimateVariable extends RasterPrecalculatedClimateVariable {
...
-		const frequencyNameMap: Record<string, string> = {
-			[FrequencyType.SEASONAL]: 'seasonal',
-			[FrequencyType.MONTHLY]: 'monthly',
-		}
+		const frequencyNameMap: Record<S2DFrequencyType, string> = {
+			[S2DFrequencyTypes.SEASONAL]: 'seasonal',
+			[S2DFrequencyTypes.MONTHLY]: 'monthly',
+			[S2DFrequencyTypes.DECADAL_ANNUAL]: 'decadal',
+			[S2DFrequencyTypes.DECADAL_MAY_SEP]: 'decadal',
+			[S2DFrequencyTypes.DECADAL_NOV_MAR]: 'decadal',
```
ChaamC: I think these should be using the individual frequency names like `decadal-ann`, `decadal-nov-mar`, ... else Geoserver layers are not requested properly.

apps/src/lib/climate-variable-base.ts
renoirb: here

apps/src/lib/s2d-climate-variable.ts
```diff
@@ -116,22 +117,30 @@ class S2DClimateVariable extends RasterPrecalculatedClimateVariable {
...
+		} as const;
```
renoirb, line 135: here

apps/src/lib/s2d.ts
```diff
@@ -309,8 +384,35 @@ export const S2D_DOWNLOAD_FILENAME_MAP_FREQUENCY_TYPE: Record<
...
 	string
 > = {
-	[FrequencyTypes.MONTHLY]: 'Monthly',
-	[FrequencyTypes.SEASONAL]: 'Seasonal',
+	[S2DFrequencyTypes.MONTHLY]: 'Monthly',
+	[S2DFrequencyTypes.SEASONAL]: 'Seasonal',
+	// [S2DFrequencyTypes.DECADAL_ANNUAL]: 'decadal-ann',      // @TODO Confirm what would be the equivalent for download
+	// [S2DFrequencyTypes.DECADAL_MAY_SEP]: 'decadal-may-sep', // ^
+	// [S2DFrequencyTypes.DECADAL_NOV_MAR]: 'decadal-nov-mar', // ^
+};
```
renoirb, line 397: Here (for download, later?)

apps/src/lib/s2d.ts
```diff
@@ -328,15 +430,18 @@ export function normalizeForApiVariableId(variableId: string): string {
...
 ): string {
 	const frequencyNameMap: Record<string, string> = {
-		[FrequencyType.SEASONAL]: 'seasonal',
-		[FrequencyType.MONTHLY]: 'monthly',
+		[S2DFrequencyTypes.SEASONAL]: 'seasonal',
+		[S2DFrequencyTypes.MONTHLY]: 'monthly',
+	  [S2DFrequencyTypes.DECADAL_ANNUAL]: S2DFrequencyTypes.DECADAL_ANNUAL,
+	  [S2DFrequencyTypes.DECADAL_MAY_SEP]: S2DFrequencyTypes.DECADAL_MAY_SEP,
+	  [S2DFrequencyTypes.DECADAL_NOV_MAR]: S2DFrequencyTypes.DECADAL_NOV_MAR,
 	};
```
renoirb, line 459: Here

apps/src/types/climate-variable-interface.ts
```diff
@@ -105,8 +104,29 @@ export const FrequencyTypes = {
...
+	 */
+	DECADAL_MAY_SEP: 'decadal-may-sep',
+	/**
+	 * Decadal Frequency only taking the "cold" months of a year.
+	 *
+	 * @remark
+	 * String 'decadal-nov-mar' is also what's expected by the Geoserver API
+	 */
+	DECADAL_NOV_MAR: 'decadal-nov-mar',
 } as const;
```
renoirb, line 124: The (supposedly one) definitions

apps/src/components/fields/frequency.tsx
```diff
@@ -1,40 +1,73 @@
...
+	S2DFrequencyTypes,
+	type S2DFrequencyType,
+} from '@/types/climate-variable-interface';
+
+const formatLabelDecadalFrequencyField = (label: string): string => {
+	return sprintf(
+		__('Decadal (%s)'),
+		sprintf(
+			'%s; %s',
+			sprintf(_n('%s year', '%s years', 5), 5),
```
ChaamC: Why use singular/plural translations, when we are only using a hardcoded value of 5?
I feel like it is overcomplexified and that it would not be needed.

apps/src/lib/s2d-climate-variable.ts
```diff
@@ -81,7 +82,7 @@ class S2DClimateVariable extends RasterPrecalculatedClimateVariable {
 	}
 
 	getFrequency(): string | null {
-		return super.getFrequency() ?? FrequencyType.SEASONAL;
+		return super.getFrequency() ?? S2DFrequencyTypes.SEASONAL;
 	}
```
renoirb, line 86: Around here. To get the currently selected frequency. (Maybe)

renoirb (commented):
Places like this might need something more.

But I couldn't be sure if we were to merge this PR like that and what's ready and what's not.

apps/src/components/fields/frequency.tsx
```diff
@@ -1,40 +1,73 @@
...
+			value: S2DFrequencyTypes.MONTHLY,
+			label: __('Monthly'),
+		},
+		{
+			value: S2DFrequencyTypes.SEASONAL,
+			label: __('Seasonal (3 months)'),
+		},
+		{
+			value: S2DFrequencyTypes.DECADAL_ANNUAL,
+			label: formatLabelDecadalFrequencyField('Annual'),
```
ChaamC: For the struggles about capitalization, couldn't this have been simplified by just writing the actual string we are using, and translation that whole string in the `fr_CA.po` file afterward?

e.g. :
`label: __('Decadal (5 years; Annual)',`

Which would give us more control over the display of the french translation.

It would also be much simpler to read the code with hardcoded strings here, instead of the formatting function above that uses 3 nested `sprintf` calls.

apps/src/components/fields/frequency.tsx
```diff
@@ -43,7 +76,7 @@ export const S2DFrequencyFieldDropdown = (
 	};
 
 	return (
-		<Dropdown<FrequencyType | string>
+		<Dropdown<S2DFrequencyType | string>
 			key={FrequencyField.key}
 			placeholder={__('Select an option')}
 			options={FrequencyField.options}
```
renoirb: Feature toggle here?

Say we modify `FrequencyField.options`, remove the DECADAL ones unless a cookie exists... the "*toggle*" being by default to hide.  Until follow-up (targeted, smaller) PRs finishes things up

apps/src/components/map-layers/s2d-build-forecast-categories.ts
```diff
@@ -8,12 +8,12 @@ import {
...
 
 /**
- * Builds forecast category definitions for the given forecast type.
+ * Builds forecast category definitions for the given forecast type
  *
  * Used by both the map legend tooltip and the probability bars tooltip
  * to share consistent category names and percentile descriptions.
  */
-export const buildForecastCategories = (
+export const buildForecastCategories = ( // Suggested rename `buildForecastProbabilitiesCategories`
```
ChaamC: Remove comment

renoirb (commented):
Feature toggle?

apps/src/types/climate-variable-interface.ts
```diff
@@ -105,8 +104,29 @@ export const FrequencyTypes = {
...
 export const S2DFrequencyTypes = {
-    MONTHLY: FrequencyTypes.MONTHLY,
-    SEASONAL: FrequencyTypes.SEASONAL,
+	MONTHLY: FrequencyType.MONTHLY,
+	SEASONAL: FrequencyType.SEASONAL,
+	/**
+	 * Decadal Frequency taking all 12 months of the year into account.
+	 *
+	 * @remark
+	 * String 'decadal-ann' is also what's expected by the Geoserver API
```
ChaamC: I would remove those remarks. It is already assumed that both API and Geoserver use the same frequency type names.

apps/src/components/map-layers/s2d-variable-values.tsx
```diff
@@ -185,7 +192,7 @@ const isWhite = (colour: `#${string}`): boolean =>
  * @param colorMap - The colour map containing the colours
  * @returns The colour associated with the given percentage and outcome.
  */
-export const getProbabilityColour = (
+export const getProbabilityColour = ( // Suggested rename `getProbabilitiesBarChartColour`
```
ChaamC: I don't mind the rename, but it shouldn't stay as a comment here.

I don't mind keeping the old name either, as the function in itself doesn't necessarily deals with BarCharts, although I get that it is used in the context of the BarCharts.

The old name keeps it more neutral/decoupled, but the new name gives the context in which it is used.

apps/src/components/map-layers/s2d-variable-values.tsx
```diff
@@ -927,24 +967,44 @@ const ProbabilitiesPart = (
...
+	let DateRangeLine = dateRangeStart
 		? generatePeriodRangeLabel(dateRangeStart, frequency, locale)
 		: null;
+
+	if (isFrequencyTypeDecadal(frequency) && DateRangeLine !== null) {
+		/**
+		 * Support both 'Juillet à Août' (or other variants) and differentiate when we want '2026-2030' instead of what would be '2026 à 2030'
+		 *
+		 * rel: CLIM-1447 We need only the year when `FrequencyTypes.DECADAL_...`.
+		 * Manual confirmation made that `Intl.DateTimeFormat(locale, {...}).format()` returns the same "month year" string ordering for both French and English.
```
ChaamC: I am not sure this line of comment should be here. If required, I would document this in the actual function that uses `Intl.DateTimeFormat(...)`

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
```diff
@@ -53,22 +57,101 @@ const generateSliderLabels = (
...
 
-	const minimumLabel = formatMinMaxLabel(firstPeriod, locale);
-	const maximumLabel = formatMinMaxLabel(lastPeriod, locale);
+	let minimumLabel = formatMinMaxLabel(firstPeriod, locale);
+	let maximumLabel = formatMinMaxLabel(lastPeriod, locale);
+
+	if (isFrequencyTypeDecadal(frequencyType)) {
+		/**
+		 * rel: CLIM-1447 We need only the year when `FrequencyTypes.DECADAL_...`.
+		 * Manual confirmation made that `Intl.DateTimeFormat(locale, {...}).format()` returns the same "month year" string ordering for both French and English.
```
ChaamC: See other feedback on the same comment found in `s2d-variable-values.tsx`

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
```diff
@@ -53,22 +57,101 @@ const generateSliderLabels = (
...
 		: formatShortMonth;
 
-	const minimumLabel = formatMinMaxLabel(firstPeriod, locale);
-	const maximumLabel = formatMinMaxLabel(lastPeriod, locale);
+	let minimumLabel = formatMinMaxLabel(firstPeriod, locale);
+	let maximumLabel = formatMinMaxLabel(lastPeriod, locale);
+
+	if (isFrequencyTypeDecadal(frequencyType)) {
+		/**
+		 * rel: CLIM-1447 We need only the year when `FrequencyTypes.DECADAL_...`.
```
ChaamC: I don't feel we need to specifically mention the Jira task, but the rest of the comment is relevant.

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
```diff
@@ -260,7 +356,8 @@ const TimePeriodsControlS2D: React.FC<TimePeriodsControlS2DProps> = ({
 				<div
 					className={cn(
 						'flex justify-between mt-2.5 mx-4 text-sm uppercase',
-						isLoadingReleaseDate && 'hidden'
+						isLoadingReleaseDate && 'hidden',
+						shouldSliderBeDisabled && 'opacity-50',
```
ChaamC: Do you think we could just hide the time period section completely for climatology data + decadal frequency, instead of greying it out (aka changing opacity)?

Might be cleaner, especially since it currently displays some year values, which are not relevant for climatology.

(see the 2026-2030 value displayed in the image)
Image

ChaamC: Also, the time periods (e.g. 2026-2030) should be hidden in the location popup too, for the climatology/decadal combination.

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
```diff
@@ -196,6 +289,8 @@ const TimePeriodsControlS2D: React.FC<TimePeriodsControlS2DProps> = ({
 		setDateRange(formatPeriodRange(period));
 	};
 
+	// @TODO Check what to adjust `ControlTitle` and `Slider.Root` when Climatology + Decadal
+	//       `forecastDisplay !== ForecastDisplays.CLIMATOLOGY`
```
ChaamC: Should this comment be removed?

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
```diff
@@ -53,22 +57,101 @@ const generateSliderLabels = (
...
+			 * const tickLabels = [
+			 *   "2026-2030",
+			 *   "2031-2035"
+			 * ]
+			 * ```
+			 *
+			 * Notice:
+			 * - No need to fence for `ForecastDisplays.FORECAST` since a DECADAL is what makes the difference for this.
+			 * - Only 2 items
+			 */
```
ChaamC: Feels too verbose.
Contains a TODO?

I would debate the comment. I think it should document/clarify if needed, for future development concerns.
Same thing for the similar comment in the `else` below.
Can't they be summarized in 1-2 lines just to clarify the expected produced output for example?

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
```diff
@@ -53,22 +57,101 @@ const generateSliderLabels = (
...
+	 * What is now `isActuallyForecast` was originally `showYear` which would make `minimumLabel` have the year.
+	 *
+	 * ```js
+	 * const sliderLabels = {
+	 *   minimumLabel: 'août 2026',
+	 *   maximumLabel: 'juil. 2027',
+	 *   tickLabels: [ ... ]
+	 * }
+	 * ```
+	 */
```
ChaamC: Similarly as above comments, review this documentation.
No need to document what was historically done.

apps/src/lib/s2d.ts
```diff
@@ -47,44 +46,101 @@ export interface LocationS2DData {
...
+	throws: boolean = true,
+): [number, PeriodJumpUnit] | null => {
+	let periodJumpSize = 1;
+	let periodJumpSizeUnit: PeriodJumpUnit = 'month';
+	if (frequency === S2DFrequencyTypes.SEASONAL) {
+		// Because a season is 3 months long.
+		periodJumpSize = 3;
+	} else if (isFrequencyTypeDecadal(frequency)) {
+		periodJumpSizeUnit = 'year';
+		periodJumpSize = 5; // There might be some logic to adjust this here
```
ChaamC: Comment doesn't seem useful in the long term for documentation

apps/src/lib/s2d.ts
```diff
@@ -328,15 +430,18 @@ export function normalizeForApiVariableId(variableId: string): string {
...
  * @returns The frequency name to use for S2D API requests, e.g. `"seasonal"`
  */
 export function normalizeForApiFrequencyName(
 	frequency: S2DFrequencyType | string
 ): string {
 	const frequencyNameMap: Record<string, string> = {
-		[FrequencyType.SEASONAL]: 'seasonal',
-		[FrequencyType.MONTHLY]: 'monthly',
+		[S2DFrequencyTypes.SEASONAL]: 'seasonal',
+		[S2DFrequencyTypes.MONTHLY]: 'monthly',
```
ChaamC, line 455: I was wondering why those are not using `S2DFrequencyTypes.SEASONAL`/`S2DFrequencyTypes.MONTHLY`, but it seems a bit messy, as they use `FrequencyType.MONTHLY`/`FrequencyType.SEASONAL`, which seem to be used for other data outside of S2D too.

Commenting here to note it, but we might want to keep this outside this PR, since this could add lots of noise to the Decadal feature.

OK to leave it as is for now.

apps/src/lib/s2d.test.ts
```diff
@@ -111,6 +122,82 @@ describe('getPeriods', () => {
 		}
 	);
 
+	describe('Confirm documented behavior', () => {
+		const releaseDate = utc('2025-08-05') as Date;
```
ChaamC: Just a note, about decadal frequencies, in the real-world application, they would never technically have a release date that is not on January 1st.
Maybe add a short comment to clarify this?

apps/src/components/fields/frequency.tsx
ChaamC: This seems to make sense as a place for a feature toggle.

apps/src/components/fields/frequency.tsx
renoirb: That's because `_n('key when singular', 'key when plural', N)` is the way to make the i18n framework know which key to use. We know it'll ALWAYS BE 5 years. But. That way we can use the same '`Decadal (%s)`' as elsewhere.

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
renoirb: That's for me next PR to have tests to actually have these situations codified as tests... instead of checking by clicking things. That will be rewritten and removed in the follow-up #712. Something similar to what I did for making sure how true the comments (see [here](https://github.com/CanadianClimateDataPortal/climatedata-wp-theme/pull/712#pullrequestreview-4996905823) in 712)

apps/src/lib/s2d.ts
renoirb: Ah. I wrote "there might b esome..." when I wasn't sure that this value would actually do the behavior we wanted. It does.

Easy to remove.

apps/src/lib/s2d-climate-variable.ts
renoirb: true, that's most likely one of the places we have to adjust like you're saying.

apps/src/lib/s2d.ts
renoirb: This was a reminder .. because when I wrote it, we were focussing on having things be displayed and organized locally. 

Not to also include working correctly with the backend that was still in *flux*.

apps/src/lib/s2d.test.ts
renoirb: > Just a note, about decadal frequencies, in the real-world application, they would never technically have a release date that is not on January 1st.

Ah. 

I could never know that.

So. When does this happen?  Only on decadal `S2DFrequencyTypes.DECADAL_*` frequencies?  Because S2D also has the word "decadal" (e.g. S2D....monthly). Just to confirm when you say "Decadal frequency" you mean the ones with ALSO the word decadal in the name

ChaamC: Well, it's all decadal frequencies, so the 3 decadal frequencies `decadal-ann`, `decadal-may-sep` and `decadal-nov-mar`.

But, of course, the release date is handled by the API server, so it shouldn't be too much of a concern. It was just that in this test it is using another release date value, which is not necessarily representative of the ones usually returned by the API. So, either we use another testing date value for decadal frequencies that falls on the January 1st for this test, or we just add a comment mentioning that decadal usually have a release date falling on January 1st, but that we are using an arbitrary value here for testing.

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
renoirb: That's for giving a relationship or subject for the following as it's on the same subject. That's only for the transition between this PR and the follow-up #712.

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
renoirb: yeah. It's too much complexity.  I dislike making things "*jump around*" and change because elements gets removed around. But that's what's going on at the moment anyway.

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
renoirb: Yes. This is part of the slider and when I was trying to keep it (so things don't "*jump around*")

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
renoirb: The issue are the tests (or lack of). But that's just a needle in a haystack, within a *black hole* (like, a very small concern) of other things we could do.

apps/src/lib/s2d.ts
renoirb: True. Anyway. Now with this, we get something that speaks by itself: 

```js
const [size, unit] = resolveFrequencyPeriodJump(S2DFrequencyTypes.DECADAL_ANNUAL);
// [5, 'year']
```

Which was my goal to reach something more self descriptive.

apps/src/lib/s2d.ts
renoirb: I know. It itches me too.

If we have a "registry" of ALL possible frequencies. That would be `FrequencyTypes`. But at the moment, we've treated `FrequencyTypes` BOTH for that "registry" and lots of conditional checks.

I wanted to keep the symbol for `FrequencyTypes.SEASONAL` to be exactly the same as any other instance so we don't break when the check is on Seasonal. But we have a very specific type "S2DFrequency" that can't EVER exist as `FrequencyType` OUTSIDE of S2D.

So. `S2DFrequencyTypes` contains a join of only two generic frequencies AND what's unique to S2d.

apps/src/components/map-layers/s2d-variable-values.tsx
```diff
@@ -927,24 +967,55 @@ const ProbabilitiesPart = (
...
+	if (!isForecast && isFrequencyTypeDecadal(frequency)) {
+		/**
+		 * Climatology uses the same data for every time period, so a period
+		 * such as '2026-2030' would suggest the values belong to those years.
+		 *
+		 * We're Removing it to not cause confusion.
+		 * That will make a floating "Décennale (...)" sub-heading in smaller text.
+		 */
+		DateRangeLine = null;
+	}
```
renoirb: This line will do this instead:

Image

(**NOTHING HERE** is my own addition, there won't be anything)

apps/src/components/fields/frequency.tsx
```diff
@@ -43,13 +106,13 @@ export const S2DFrequencyFieldDropdown = (
 	};
 
 	return (
-		<Dropdown<FrequencyType | string>
+		<Dropdown<S2DFrequencyType | string>
 			key={FrequencyField.key}
 			placeholder={__('Select an option')}
-			options={FrequencyField.options}
+			options={options}
```
renoirb: Here's how we can have the frequencies added behind a feature-toggle.

apps/src/lib/s2d.test.ts
renoirb: Added comment "For `S2DFrequencyTypes.DECADAL_*` frequencies, the `releaseDate` would ALWAYS be set to January the 1st of the year."

apps/src/components/sidebar-menu-items/time-periods-control-s2d.tsx
ChaamC: A worry I have is that the '2026-2030' is still displayed (even if greyd out), but it is not related to Climatology data, since climatology data is based on 1991-2020 average data. Hence why it could have been better to hide it. 

Would it be complex to hide it? I feel like it might not have that much impact on the UI elements, since there is already variations in the displayed elements on the left panel on the Maps page, depending of the datasets and variables.

apps/src/lib/feature-toggling/cookie-jar.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Turns `a=1; b=2` into `Map { 'a' => '1', 'b' => '2' }`.
+ *
+ * This function is pure, so it can be tested without a browser or framework.
+ *
+ * Pairs without `=` are skipped.
+ * Both sides are trimmed because cookie entries are separated by `'; '`.
+ */
+export const parseCookieString = (raw: string): CookieEntries => {
+	// Illustration.
```
ChaamC: Why all those "Illustration" mention in this file and the `cookie-toggles.ts` file?

apps/src/components/fields/frequency.tsx
renoirb: I've removed this one. This was just too much.

apps/src/lib/feature-toggling/cookie-jar.ts
renoirb: They were made when I was writing the first implementation and I wanted to use as "illustration" that I had said in the earlier meeting when we'd discussed about them. I don't mind or need these comments.

I've removed them.

fw-child/languages/react-apps/fr_CA.po
```diff
@@ -1210,6 +1210,9 @@ msgstr "Mai-Sep"
 msgid  "Nov-Mar"
 msgstr "Nov-Mar"
 
+msgid "Unsupported frequency type: %s"
+msgstr "Type de fréquence non pris en charge : %s"
```
ChaamC, line 1214: Should we really translate an exception message? This is not displayed on the UI?

apps/src/components/map-layers/location-modal-content.tsx
```diff
@@ -78,6 +78,7 @@ export const LocationModalContent: React.FC<LocationModalContentProps> = ({
 		);
 	}
 
+	// The `data-part` attributes below, are inspection landmarks.
```
ChaamC, line 81: What do you mean by "inspection landmarks"?
Not sure about this comment.

apps/src/lib/raster-precalculated-climate-variable.tsx
```diff
@@ -240,8 +240,9 @@ class RasterPrecalculatedClimateVariable extends ClimateVariableBase {
 			if (payload.month && payload.dataset_name) {
 				const frequencyType = getFrequencyType(payload.month);
 				const scenario = payload.dataset_name;
-				// S2D frequencies have no preCalculatedCanDCSConfig entries. Note
-				// S2DFrequencyTypes covers MONTHLY and SEASONAL, not only decadal.
+				// CanDCS has no pre-calculated entry for any S2D frequency, and
+				// `getFrequencyType` now returns S2D codes rather than `undefined`.
+				// The guard covers the whole S2D set, monthly and seasonal included,
```
ChaamC: Not sure about the second part of this comment. 
It references a previous version of the code? It might not be relevant.

And it is confusing if reading this out of context.
`getFrequencyType` can return all types of frequencies, not juste S2D codes, but this comment can make it seem as if it just returns s2d codes.

I think just the first part is relevant `CanDCS has no pre-calculated entry for any S2D frequency.`

apps/src/lib/feature-toggling/cookie-writes.ts
```diff
@@ -0,0 +1,48 @@
...
+const TOGGLE_COOKIE_LIFETIME_IN_SECONDS = 60 * 60 * 24 * 90; // 90 days.
+
+// The `typeof` guard keeps this module usable in unit tests without a DOM.
+const isDocumentAvailable = (): boolean => typeof document !== 'undefined';
+
+/**
+ * Sets the toggle cookie, so `hasCookie(name)` returns `true` from now on.
+ *
+ * The value is `yes`, which is what someone entering the cookie by hand types.
+ * Nothing reads the value.
```
ChaamC: Maybe confusing, since some other functions do read the values (e.g. `isCookieTrue()` and `isCookieFalse()`).

renoirb: Maybe it should be something like ... an unicode symbol. Or some other word. Maybe that's what you're asking about?

---

That's because this implementation of feature flagging supports 3 possibilities. But we use only one of them. Only IF the cookie exist at all.

The 3 possibilities is that we can support other use-cases later and have only to adjust the other names and the mechanism to check can be already in place.

- The cookie with a predetermined name exists or it doesn't.
  - When it exists.
    - It exists. We only need to know if it exists. (what we're using)
    - It exists AND its value is EITHER intended to mean a boolean type value: `true`, and `false`. Nothing else.
      - It exists, AND its value is intended to mean `true`
      - It exists, AND its value is intended to mean `false`

With these 3 kind of values, we can give any feature a toggle name with  any wording and pick one of these possibilities.

apps/src/components/map-layers/location-modal-content.tsx
renoirb: @ChaamC "inspection landmarks". So you can check the DevTools and find "`location-modal-content-title`" (as an example here)  and in the code.  It was hard to find what component was responsible of that part of the page.   There's no benefit of making it "obscure" and hard to find. Particularly if you try to see what conditions make a piece display or another

PR #712, opened 2026-08-20
CLIM-1447 (2): S2D-Maps Add Decadal Time Periods Display Support

renoirb:
## Description

The last commits of #711 were to undo refactor to be applied here.

## Related ticket
- CLIM-1447
- Pass 1: #711


apps/src/lib/s2d/calculate-periods.ts
```diff
@@ -0,0 +1,254 @@
...
+};
+
+
+/**
+ * Return the time periods for a release date and specific frequency.
+ *
+ * The number and length of time periods depend on the frequency:
+ * - Seasonal frequency: 10 x 3-month periods, at 1-month interval
+ * - Monthly frequency: 3 x 1-month periods, at 1-month interval
+ * - Decadal frequency: 2 x 5-year periods, at 5-year interval
```
renoirb, line 98: The statements:  "`A` frequency: `B` x `C`-`D` periods, at `E`-`F` interval"  and making sure the code actually does them.

apps/src/lib/s2d/calculate-periods.test.ts
renoirb, line 168: To ensure the comments matches the code.

renoirb (commented):
That's the kind of tests I've added to make sure we have 1-1 relationship between what we documented.

ChaamC (approved):
Approved!
Seems to be a good restructuration of the code.
Tested locally too and seems good.

